PR #1, opened 2024-05-31
created project template

kialj876:
https://github.com/bcgov/entity/issues/21296

Will need to verify the CI/CD after this is merged, but tests / lint pass locally. Still need to sort out temp dev/test and PR builds with patrick wei


kialj876:
Local run look:
Screenshot 2024-06-03 at 11 51 49 AM

README.md
```diff
@@ -1,2 +1,66 @@
...
+Locally preview production build:
+
+```bash
+# pnpm
+pnpm run preview
+```
+
+Check out the [deployment documentation](https://nuxt.com/docs/getting-started/deployment) for more information.
+
+# Proposed Technical Plan for Entity UIs
```
hfekete, line 44: Maybe update this to say that this is specific UI rather then plan of entire entity stuff ?

package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@mdi/font": "^7.4.47",
+    "@nuxt/ui": "^2.16.0",
+    "@nuxtjs/i18n": "^8.3.1",
+    "@pinia/nuxt": "^0.5.1",
+    "@vuepic/vue-datepicker": "^8.7.0",
+    "@vueuse/core": "^10.10.0",
+    "http-status-codes": "^2.3.0",
+    "keycloak-js": "^24.0.4",
+    "launchdarkly-vue-client-sdk": "^2.2.0",
+    "lodash": "^4.17.21",
```
hfekete, line 32: what is the reason to seting it on this version, dont remember seeing this in our other json files.

src/app.vue
```diff
@@ -0,0 +1,100 @@
...
+      <base-dialog
+        id="error-dialog"
+        attach="#appHeader"
+        :display="errorDisplay"
+        :options="errorInfo"
+        @close="clearDialog"
+      >
+        <template v-if="errorContactInfo" #extra-content>
+          <p class="font-normal mt-7">
+            If this issue persists, please contact us.
```
hfekete, line 13: should we get this into language files

src/components/base/dialog/Index.vue
```diff
@@ -0,0 +1,69 @@
...
+  </UModal>
+</template>
+
+<script setup lang="ts">
+import { DialogButton, DialogContent } from './slot-templates'
+
+const props = defineProps<{
+  attach?: string,
+  display: boolean,
+  options: DialogOptionsI | null
```
hfekete, line 55: I would potentially bring in this Interface into this file, or even have inline declaration if its not to big. As I dont think anything else will use this interface ?

hfekete, line 55: Additional thing, I would like that we use undefined when variable is unset. And use null only in situation when we want to do some specific stuff like clear existing thing and make sure that everyone knows it is not never defined, but purposely put to nothing.

This is good for when we are using optional stuff in schema, they are all undefined, and not null.

src/composables/useDatetime.ts
```diff
@@ -0,0 +1,15 @@
...
+export const useDatetime = () => {
+  const dateTimeString = (val: string): string => {
+    return (dateToPacificDateTimeShort(new Date(val)) || 'Unknown')
+  }
+  const pacificDate = (date: Date, format = 'MMMM D, Y') => {
+    date = new Date(date.toLocaleString('en-US', { timeZone: 'America/Vancouver' }))
+    return moment(date).format(format)
+  }
+  return {
+    dateTimeString,
```
hfekete, line 12: can we make this into iso date format ?

src/lang/en.json
```diff
@@ -0,0 +1,59 @@
...
+        "createAccount": "Create Account",
+        "editProfile": "Edit Profile",
+        "logOut": "Log Out",
+        "teamMembers": "Team Members",
+        "transactions": "Transactions"
+      }
+    }
+  },
+  "footer": {
+    "labels": {
```
hfekete, line 52: I think we are mixing again.
we have 
`labels -> general -> email`
and we have
`footer -> labels -> ....`

src/resources/corp-info.ts
```diff
@@ -0,0 +1,294 @@
...
+    shortDesc: 'XPRO LIM PARTNR',
+    fullDesc: 'Extraprovincial Limited Partnership'
+  }, {
+    corpTypeCd: CorpTypeCdE.XPRO_SOCIETY,
+    colinInd: true,
+    corpClass: CorpClassE.SOC,
+    shortDesc: 'XPRO SOCIETY',
+    fullDesc: 'Extraprovincial Society'
+  }
+]
```
hfekete, line 294: Haha, hope this was copy pastable from somewhere :D

src/utils/date.ts
```diff
@@ -0,0 +1,119 @@
...
+import { isDate } from 'lodash'
+
+/**
+   * Converts a Date object to a date string (Month Day, Year) in Pacific timezone.
+   * @param longMonth whether to show long month name (eg, December vs Dec)
+   * @param showWeekday whether to show the weekday name (eg, Thursday)
+   * @example "2021-01-01 07:00:00 GMT" -> "Dec 31, 2020"
+   * @example "2021-01-01 08:00:00 GMT" -> "Jan 1, 2021"
+   */
+export function dateToPacificDate (date: Date, longMonth = false, showWeekday = false): string {
```
hfekete, line 11: do we need this file if we have composable ?

hfekete (commented):
Awesome work, I see some things are cleaned up very nicely !
Just had few questions

README.md
kialj876: Good idea will update

package.json
kialj876: this is just 4.17.21 or higher. It is set the same way in director search and should be the latest version

src/app.vue
kialj876: yep good point

src/components/base/dialog/Index.vue
kialj876: it used elsewhere for anything that sets dialog options (ie. resource/util files for options and anywhere initializing the dialog like app.vue)

kialj876: Good point, will update

src/lang/en.json
kialj876: right -- was thinking we would shift some things logically for the header/footer so that its contained a bit differently and doesn't ever conflict with the project at hand (for when we import these from layers). This would mean some things would have their own section so they are always separate and never conflict when the en.json is merged. That was my thinking here

src/resources/corp-info.ts
kialj876: yes haha

src/composables/useDatetime.ts
kialj876: removed this composable (replaced by same date util that we have in BTR)

src/lang/en.json
kialj876: updated en.json as discussed

src/utils/date.ts
kialj876: replaced this and the composable with the util we have in BTR

package.json
kialj876: removed lodash entirely (we don't need it now)

README.md
kialj876: removed this entirely as its not really the right place for it. Tracking this with the entity team in this ticket: https://github.com/bcgov/entity/issues/21469

hfekete (approved):
Awesome stuff ! Thanks for updates !

PR #4, opened 2024-06-05
minor cleanup in prep for consolidation across projects

kialj876:
*Issue:*https://github.com/bcgov/entity/issues/21296

*Description of changes:*

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -59,7 +59,7 @@ export default defineNuxtConfig({
       registryHomeURL: process.env.VUE_APP_REGISTRY_HOME_URL || '',
       appEnv: `${process.env.VUE_APP_POD_NAMESPACE || 'unknown'}`,
       requireLogin: true,
-      version: process.env.npm_package_version || '',
+      version: `BCROS Business Dashboard UI v${process.env.npm_package_version || ''}`,
       appName: process.env.npm_package_name || ''
     }
   }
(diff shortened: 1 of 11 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-4-uvkzqxys.web.app

nuxt.config.ts
```diff
@@ -59,7 +59,7 @@ export default defineNuxtConfig({
       registryHomeURL: process.env.VUE_APP_REGISTRY_HOME_URL || '',
       appEnv: `${process.env.VUE_APP_POD_NAMESPACE || 'unknown'}`,
       requireLogin: true,
-      version: process.env.npm_package_version || '',
+      version: `BCROS Business Dashboard UI v${process.env.npm_package_version || ''}`,
```
hfekete, line 62: Do we want to put text in the version ?
Would it make sense to add additional variable and call it something like App 
and have App + Version displayed ?

hfekete: It could be even in language file, then App name would be translated too if needed.

hfekete (approved):
Looks good, just one question about app name. If you think this is good for now, its good to merge!

nuxt.config.ts
kialj876: Sure I'll set it from the config as a different variable. I don't think it makes sense in the language file though because it is a name not a description

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-4-uvkzqxys.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-4-uvkzqxys.web.app

PR #12, opened 2024-06-27
Alerts

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/21301

*Description of changes:*

This adds the alerts and covers the issue #21301

It is dependent on the changes to the business store that add access to all of the api values


src/components/bcros/Alert.vue
```diff
@@ -0,0 +1,103 @@
+<template>
```
kialj876, line 1: I think its worth putting this inside 'bcros/alert/Index.vue' and then having the 'Alerts.vue' -> 'bcros/alert/List.vue' since they are sort of grouped together. What do you think?

src/components/bcros/Alert.vue
```diff
@@ -0,0 +1,103 @@
...
+  <div data-cy="alert-display" class="px-3 py-3">
+    <h3>
+      <UIcon
+        :class="`${iconColour} mr-2`"
+        :name="iconName"
+        data-cy="contact-icon"
+      />
+      <span class="font-semibold flex-auto">{{$t(alertHeader)}}</span>
+      <button class="float-right text-blue-600 hover:bg-blue-100 text-sm px-1 py-1mr-4" href="#" @click="expanded = !expanded">
```
kialj876, line 10: can you replace 'text-blue' with 'text-primary'? It will evaluate the same right now, but if we ever change our primary colour in the future we would want this to change with it. Also, did 600 come from the UX or should that be '500'? I think we are using '500' everywhere else for these sort of things

kialj876 (commented):
Hey Brandon, nice work this is looking good! I just have a couple minor comments and can you please add a cypress test / fix the lint errors? Also I am assuming you need Hrvoje's PR to go through with the store changes before we can run the PR build?

src/components/bcros/Alert.vue
hfekete: Agree with this, I think we are doing most of the stuff like that.
Also, I would maybe call it 
BusinessAlerts or DashboardAlerts or smth like that rather then just alerts, it could clash the naming with other types of alerts.
@kialj876 what do you think about naming it bit different ?

src/enums/warning-types.ts
```diff
@@ -0,0 +1,7 @@
+export enum WarningTypes {
```
hfekete, line 1: we have file naming and Enum types by appending letter E at the end.

patrickpeinanw:
If there are multiple alerts, when one alert is expanded, top padding and bottom padding are added to this alert so it is separated from others. I guess we don't have many example with multiple alerts, but you can refer to the 'Recent Filing History' section which has the same behavior.

src/components/bcros/Alert.vue
```diff
@@ -0,0 +1,103 @@
...
+        {{ !expanded ? 'View Details' : 'Hide Details' }}
+        <UIcon
+          :class="`text-blue-600 mr-2`"
+          :name="!expanded ? 'i-mdi-chevron-down' : 'i-mdi-chevron-up'"
+          data-cy="contact-icon"
+        />
+      </button>
+    </h3>
+    <div v-if="expanded">
```
kialj876, line 19: Also is there a reason you didn't use the UAccordian for the alerts? That might simplify things in here quite a bit

src/components/bcros/Alert.vue
BrandonSharratt: This component is more generic than that, it can also take severity, text and description and then it works as a regular alert, the business specific stuff is in dashboard and alertType: frozen for instance

src/components/bcros/Alert.vue
BrandonSharratt: No particular reason for the 600 will change

src/enums/warning-types.ts
BrandonSharratt: I did just missed it here thanks

src/components/bcros/Alert.vue
kialj876: Yeah you're right and we can add a slot over the contact info if we want it to handle something else later. I am good with it staying as BcrosAlert

src/components/bcros/Alert.vue
kialj876: discussed in chat to keep it as is for now and update later to use code that is still in progress

src/components/bcros/Alert.vue
hfekete: OK, like we discussed in chat, we can update naming in the future.
Also, frozen sound like business status, so I would try to keep that out of alert itself, and let parent component handle how to display it.

hfekete: Either way, I am fine for this PR as is, so you can resolve this comment.

kialj876: yeah it looks like the other 'alert types' in the enum are business specific too, but we can update the naming later

src/components/bcros/Alert.vue
BrandonSharratt: Fixed

BrandonSharratt:
> If there are multiple alerts, when one alert is expanded, top padding and bottom padding are added to this alert so it is separated from others. I guess we don't have many example with multiple alerts, but you can refer to the 'Recent Filing History' section which has the same behavior.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-12-r41fmuic.web.app

src/components/bcros/alert/index.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <div :data-cy="`alert-display${index || index===0 ? '-' + index : ''}`" class="px-3 py-3">
+    <UIcon
+      v-if="showHeader"
+      :class="`${iconColour} mr-2 font-semibold`"
+      :name="iconName"
+      data-cy="alert-icon"
+    />
+    <span v-if="showHeader" class="font-semibold flex-auto">{{ alertHeader }}</span>
+    <button
```
kialj876, line 10: Hey Brandon can you update this to use UButton so that it picks up our common styling etc.? You will be able to remove the UIcon and it will size it similar to our other ones.
i.e. for this example it would be:
```
    <UButton
        color="primary"
        icon="i-mdi-chevron-down"
        :label="expanded ? 'Hide Details' : 'View Details'"
        trailing
        variant="ghost"
        :ui="{ icon: { base: expanded ? 'rotate-180 transition-all' : 'transition-all' } }"
        @click="expanded = !expanded"
      />
```

src/components/bcros/alert/index.vue
```diff
@@ -0,0 +1,94 @@
...
+    >
+      {{ !expanded ? 'View Details' : 'Hide Details' }}
+      <UIcon
+        v-if="showHeader"
+        class="font-semibold text-primary-500 mr-2"
+        :name="!expanded ? 'i-mdi-chevron-down' : 'i-mdi-chevron-up'"
+        data-cy="expand-icon"
+      />
+    </button>
+    <div v-if="expanded && showDescription" :data-cy="`alert-description${index || index===0 ? '-' + index : ''}`">
```
kialj876, line 24: I think you can do away with the `${index || index===0 ? '-' + index : ''}` logic here and above and just reference the index you want in your tests right?

i.e. '<selector>.eq(index)' inside the tests and then you don't need the index at all?

example of changes to test:
```
cy.get('[data-cy="alerts-display"]').should('exist')
cy.get('[data-cy="alerts-display"]').should('have.length', 2)
cy.get('[data-cy="alert-display"]').eq(0).should('contain.text', 'This business is frozen')
...
```

src/interfaces/alert-i.ts
```diff
@@ -0,0 +1,10 @@
+import type { AlertTypesE } from '~/enums/alert-types-e'
+import type { AlertSeverityE } from '~/enums/alert-severity-e'
+
+export interface AlertI {
+  severity: AlertSeverityE | null
```
kialj876, line 5: are these values ever null? Or are they just undefined sometimes?

kialj876 (commented):
Nice work brandon, I just have a couple more minor comments

src/interfaces/alert-i.ts
kialj876: also in general I think it is better to import enums/interfaces from '#imports' so that we don't have to update as many places if we change the location of things later (i.e. move some enum stuff to a common repo or rename the file etc.)

patrickpeinanw:
image
It seems that the alert section will push down the side components.

src/components/bcros/alert/index.vue
BrandonSharratt: Done this took a few additional changes but I believe this has better functionality now than before

src/components/bcros/alert/index.vue
BrandonSharratt: Done

src/interfaces/alert-i.ts
BrandonSharratt: Changed to undefined and to #imports

BrandonSharratt:
> image It seems that the alert section will push down the side components.

Thanks, fixed

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-12-r41fmuic.web.app

src/interfaces/alert-i.ts
```diff
@@ -0,0 +1,9 @@
+import type { AlertTypesE, AlertSeverityE } from '#imports'
+
+export interface AlertI {
+  severity: AlertSeverityE | undefined
+  alertType: AlertTypesE | undefined
+  text: string | undefined
+  description: string | undefined
+  date: any | undefined
```
kialj876, line 8: Note for future: shorthand would be like *severity?: AlertSeverityE* , but I think you reference this interface with Partial anyways so I don't think you need to specify the undefined possibility here

kialj876 (approved):
Nice work! Thanks for all the updates. Merge away

PR #16, opened 2024-07-11
21297 - Fix for name for soleprop, fix for view and change address link

hfekete:
*Issue:21297 *https://github.com/bcgov/entity/issues/21297 

*Description of changes:*
- fix for Sole Prop name not being displayed, and subtitle text.
- fix for link "View and Change Business Information" disabled state and opening action 


src/lang/en.json
```diff
@@ -59,7 +59,7 @@
         "BEN": "BC Benefit Company",
         "BC": "BC Limited Company",
         "CP": "BC Cooperative Association",
-        "SP": "Sole Proprietorship",
+        "SP": "BC Sole Proprietorship",
```
hfekete, line 62: @kialj876 do you know what is proper translation for this, is it with BC, as is now in prod, or without ?
or should I bring it up with business ppl ?

kialj876: I think with 'BC' is correct

src/stores/business.ts
```diff
@@ -15,7 +15,7 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
   const currentBusinessIdentifier = computed((): string => currentBusiness.value.identifier)
   const currentBusinessName = computed((): string => {
     if (currentBusiness.value.alternateNames && currentBusiness.value.legalType === CorpTypeCd.SOLE_PROP) {
-      return currentBusiness.value.alternateNames[0].operatingName
+      return currentBusiness.value.alternateNames[0].name
```
kialj876, line 18: This actually needs to be updated to be: `currentBusiness.value.alternateNames.find(name => name.identifier === currentBusinessIdentifier.value) || currentBusiness.value.legalName` because there are cases of multiple 'alternate names' and the correct one for 'operating name' has the matching identifier

Also it needs to do this for GPs as well

src/interfaces/business-i.ts
```diff
@@ -87,3 +87,45 @@ export interface BusinessI extends SlimBusinessI {
   submitter?: string // not used
   warnings: Array<BusinessWarningI>
 }
+
+export const getDefaultEmptyBusiness = (): BusinessI => {
```
kialj876, line 91: I don't think this should be inside the interfaces and it isn't being used by anything can you remove ?

.env.example
```diff
@@ -17,6 +17,7 @@ VUE_APP_LEGAL_API_URL="https://legal-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_LEGAL_API_VERSION_2="/api/v2"
 VUE_APP_STATUS_API_URL="https://status-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_STATUS_API_VERSION="/api/v1"
+VUE_APP_BUSINESS_EDIT_URL=
```
kialj876, line 20: can you add the dev url here so that the pr build picks it up? https://dev.edit.business.bcregistry.gov.bc.ca/

kialj876 (commented):
Just a few comments

src/stores/business.ts
hfekete: Updated, whats the reason why would other alternate names have different identifiers ?

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-16-ey0vryfg.web.app

src/stores/business.ts
kialj876: I don't really know all the rules surrounding it, but there are many cases (went through this with Argus a couple weeks ago before they released it into prod and he said I could always find the right one by the identifier match)

kialj876: Just need to add this for GPs as well

kialj876: oh nvm you added it!

kialj876 (approved):
Looks good!

PR #17, opened 2024-07-16
To-do Action - File Annual Report

patrickpeinanw:
*Issue:*https://github.com/bcgov/entity/issues/21350

*Description of changes:*


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-17-cc7zmt6k.web.app

src/stores/business.ts
```diff
@@ -213,6 +214,270 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
     return !!requestedFiling
   }
 
+  /** Check if the specified action is allowed, else False */
+  const isAllowed = (action: AllowableActionE): boolean => {
```
hfekete, line 218: Is tis supposed to be in here or inside some util file ? I think we said we gona keep it in enumutils.ts ?

src/stores/business.ts
```diff
@@ -213,6 +214,270 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+    // const isBusiness = !!sessionStorage.getItem('BUSINESS_ID') // ie, not a temporary business
+
+    // TO-DO: the above line is commented out because we do not have 'BUSINESS_ID' in the sessionStorage
+    // For now, we check if the currentBusiness exists in the business store.
+    const isBusiness = !!useBcrosBusiness().currentBusiness.identifier
```
hfekete, line 223: If we are keeping it here, I think you dont have to use Business store, as you are inside, am not sure what are consequences of that.

src/utils/todo.ts
```diff
@@ -23,6 +26,190 @@ export const buildTodoItemIfFromAffiliationInvitation =
...
+  if (business && header) {
+    const ARFilingYear = header.ARFilingYear
+
+    const subtitle = (task.enabled && !business.isBaseCompany())
+      ? '(including Address and/or Director Change)'
+      : null
+
+    const enabled = task.enabled && !business.isDisableNonBenCorps()
+
+    // TO-DO: business.isAllowed() will always return false for KIAL DEV 1 account
```
hfekete, line 77: If not already created, please create a ticket for this issue so we look into it.

hfekete (commented):
Looks good, few comments, few questions. Also, if you did not see, some cypress are failing.

src/components/bcros/todo/List.vue
```diff
@@ -1,14 +1,20 @@
 <template>
   <div
-    class="flex flex-col gap-1.5 bg-gray-100"
+    class="flex flex-col"
```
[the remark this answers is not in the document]
patrickpeinanw: We don't need to use gap here. An extra space will be added above each todo title except for the first one. To better illustrate this, I used `gap-10` in the following screenshot. 
image

The reason why I removed `bg-gray-100` is that, when an to-do item is expanded, no top and bottom gap are added. (Alerts and Recent Filing History has such behavior but it is not the case in To-Do section)

src/interfaces/amalgamation-i.ts
```diff
@@ -0,0 +1,66 @@
+import { CorpTypeCd, AmalgamationTypes, CorrectNameOptions, FilingTypes } from '@bcrs-shared-components/enums'
+import type {
+  CompletingPartyIF, ContactPointIF, CourtOrderIF, NameTranslationIF, ShareClassIF
+} from '@bcrs-shared-components/interfaces'
+
+/**
+ * A filing's amalgamation application object from the API. See:
+ * https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/amalgamation_application.json
+ */
+export interface RegisteredRecordsAddressesI {
```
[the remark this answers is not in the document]
patrickpeinanw: I actually don't know what that is haha. Just copied from the old codebase.

src/stores/business.ts
patrickpeinanw: I added a `enum-utilities.ts` that contains some helper functions to compare filling status. (BTW, it is a `EnumUtilities` class with static methods. But maybe we can just use functions instead of putting them inside a class)

For this `isAllowed` function, it is a wrapper for `isAllowedToFile()` so I think it is better to keep them in the same file.

src/stores/business.ts
patrickpeinanw: Good catch! I will change that.  Another update I need to make is, only one todo item can be expanded each time.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-17-cc7zmt6k.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-17-cc7zmt6k.web.app

src/components/bcros/todo/Item.vue
```diff
@@ -1,33 +1,150 @@
...
+        <UButton
+          variant="outline"
+          class="px-3 w-40 h-8"
+          data-cy="todoItem-affiliation-doNotAuthorizeButton"
+          @click="todosStore.authorize(item.affiliationInvitationDetails.id, false)"
+        >
+          <span class="w-full text-center"> {{ $t('button.todoItem.doNotAuthorize') }}</span>
+        </UButton>
+        <UButton
+          class="px-3 w-40 h-8"
```
kialj876, line 61: candidate for a common button class?
i.e. 
```
.bcros-btn-md {
@apply px-3 w-40 h-8
}

```

src/interfaces/amalgamation-i.ts
```diff
@@ -0,0 +1,66 @@
...
+  incorporationAgreement?: {
+    agreementType: string
+  }
+  // ULC only:
+  courtOrder?: CourtOrderIF
+}
+
+/** Amalgamation Application filing interface. */
+export interface AmalgamationApplicationFilingI {
+  header: {
```
kialj876, line 45: can you add a 'FilingI' with the header + business sections (these are shared across all filings) and then extend it in the new filing interfaces?

src/utils/todo.ts
```diff
@@ -23,6 +26,191 @@ export const buildTodoItemIfFromAffiliationInvitation =
...
+    }
+
+    return newTodo
+  } else {
+    console.error('ERROR - invalid header or business in todo =', todo)
+  }
+}
+
+/** TO-DO load items when the task is filing. Change it to async after implementing functions that loads data */
+const buildFilinngTodo = (task: TaskI) : TodoItemI | null => {
```
kialj876, line 143: typo

src/stores/business.ts
```diff
@@ -213,6 +214,270 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+    return currentBusiness.value.legalType === CorpTypeCd.BC_COMPANY
+  }
+
+  /** Whether the entity is a Benefit Company. */
+  function isEntityBenefitCompany (): boolean {
+    return currentBusiness.value.legalType === CorpTypeCd.BENEFIT_COMPANY
+  }
+
+  /** Whether the entity is a BC Community Contribution Company. */
+  function isEntityBcCcc (): boolean {
```
kialj876, line 423: Could we replace all these 'is...' with one function where you pass the desired legal type in? Otherwise we will end up with 100 of these and I don't think having specific ones adds very much value

kialj876: and you could make the param a list so that we could check if its in a group

src/utils/enum-utilities.ts
```diff
@@ -0,0 +1,15 @@
+import { FilingTypes } from '@bcrs-shared-components/enums'
+
+// NB: These two methods are just for getPnedingCoa() in useBcrosFilings store
+// more methods to be added. Or the EnumUtilities class can be replaced by functions
```
kialj876, line 4: I would update this comment to make these more generic if necessary (i.e. to pass in the desired filing type or filing status rather than have separate functions for every possibility). Also the name 'EnumUtilities' is a bit misleading I think - could we change that to something like 'FilingUtil' as this will always be specific to filings data I think?

kialj876: actually is there a reason this isn't in the filing store? It seems intuitive to have something like isFilingType(types: FilingTypes[]) / isStatus(status: FilingStatusE[]) in the filing store, similar to what I was asking for in the business store?

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-17-cc7zmt6k.web.app

kialj876 (approved):
Nice work 👍

PR #35, opened 2024-08-27
Add Business comments

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/21782

*Description of changes:*
Added comments button to the tombstone links, reused existing modal, added fetch and create comments for business. 


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-35-2u0u9a1u.web.app

src/components/bcros/businessDetails/Links.vue
```diff
@@ -40,6 +41,14 @@ const isChangeBusinessInfoDisabled = computed(() => {
   return !isAllowed
 })
 
+const showCommentDialog = (show?: boolean) => {
+  if (typeof show !== 'boolean') {
```
hfekete, line 45: Why this check?  Do we not want to provided data be boolean and just refuse to have reasonable behavior otherwise ?
If you want to act on other types too, I would say 0 is hide, rather then use 0 to show comment box.
Also, with this, undefined, null, etc... will also display the comment box, no ?
Maybe write it as this following:`isCommentOpen.value = !!show`
This allows falsy/truthy standard behavior from JS

src/stores/business.ts
```diff
@@ -36,6 +39,21 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+    comments.value = []
+    return await useBcrosFetch<CommentIF>(`${apiURL}/businesses/${identifier}/comments`, {})
+      .then(({ data, error }) => {
+        if (error.value || !data.value) {
+          console.warn('Error fetching comments for', identifier)
+          return null
+        }
+        comments.value = data.value.comments.map(comment => comment.comment)
+        commentsLoading.value = false
+        return comments.value
```
hfekete, line 53: I would either avoid returning anything, or use constant, and assign it on function return. This is creating confusing logic, where function has return value, but it returns value which is assigned to global ref.

hfekete: e.g.
```
 async function getBusinessComments(identifier: string) {
    commentsLoading.value = true
    return await useBcrosFetch<CommentIF>(`${apiURL}/businesses/${identifier}/comments`, {})
      .then(({ data, error }) => {
        if (error.value || !data.value) {
          console.warn('Error fetching comments for', identifier)
          return null
        }
        const comments = data.value.comments.map(comment => comment.comment)
        commentsLoading.value = false
        return comments
   })
  }
```
or just remove return

src/stores/business.ts
```diff
@@ -437,6 +456,46 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+    }
+    // post comment to API
+    const commentRes = await postComment(businessId, commentObj)
+    // flatten and sort the comments
+    if (comments.value && comments.value.length > 0) {
+      comments.value = [commentRes.comment, ...comments.value]
+    } else {
+      comments.value = [commentRes.comment]
+    }
+    return commentRes.comment
```
hfekete, line 496: why return here ? I see up that no one is using return value.

src/stores/business.ts
```diff
@@ -139,6 +157,7 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
   }
 
   async function loadBusiness (identifier: string, force = false) {
+    getBusinessComments(identifier)
```
kialj876, line 160: this should go under the cached block I think?

src/stores/business.ts
```diff
@@ -437,6 +456,46 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+    return await useBcrosFetch<{ comment: CommentIF }>(url, { method: 'POST', body: { comment } })
+      .then(({ data, error }) => {
+        if (error.value || !data.value) {
+          console.warn('postComment() error - invalid response =', error?.value)
+          throw new Error(error.value.message)
+        }
+        return data?.value
+      })
+  }
+  const createCommentBusiness = async (businessId: string, comment: string): Promise<CommentIF> => {
```
kialj876, line 476: I think this should just use the currentBusinessIdentifier? I don't think we want the possibility for accidentally posting comments to other businesses

BrandonSharratt: Done

src/stores/business.ts
BrandonSharratt: Removed return value

src/components/bcros/businessDetails/Links.vue
BrandonSharratt: This was done for consistency with HeaderActions and how I had to do it there. That said I've made the change to require boolean value (and added true where it was unspecified before) and made the change here and in the other file that was mimicing this (HeaderActions is still like this)

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-35-2u0u9a1u.web.app

kialj876 (commented):
Nice work, looking good but can you fix the layout of the buttons (this is at full width) and add a test that verifies getting the comments as expected on page load and another one that verifies posting a new comment
Screenshot 2024-08-28 at 12 49 35 PM

src/stores/business.ts
hfekete: I'll have to bother you to rename the function then :D to maybe fetchBusinessComments as get usually implies return value, sorry I didnt mentioned that previously

BrandonSharratt:
@kialj876 added and fixed (the layout was outstanding issue before on smaller screens)
@hfekete I renamed the function

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-35-2u0u9a1u.web.app

kialj876 (approved):
Nice work, thanks for the updates

PR #39, opened 2024-08-30
pending address change

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/22690

*Description of changes:*
If the business has a pending address change filing the addresses will show with the pending badge, the section will be tinted yellow and the change address button will be disabled. Tests added/modified to verify badge shows up and doesn't show up as expected

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -4,8 +4,6 @@ export default {
   theme: {
     extend: {
       colors: {
-        pending: '#ffeb3b',
-        pendingtint: '#ffeb3b1a',
         bcGovColor: {
           darkBlue: '#003366'
         },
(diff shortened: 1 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-39-dn1k5f8e.web.app

tailwind.config.ts
```diff
@@ -4,6 +4,8 @@ export default {
   theme: {
     extend: {
       colors: {
+        pending: '#ffeb3b',
```
hfekete, line 7: I would put this under bcGovColor if they are going to be specific colors.
Otherwise it looks like its special entire class when used in tailwind e.g.
`bg-pending`
vs 
`bg-bcGovColor-pending`

we also did similar in the [BTR project common tailwind config](https://github.com/bcgov/business-transparency-registry/blob/main/btr-web/btr-common-components/tailwind.config.ts):

but what would probably be the best is, if it could be shade of e.g. yellow or smth similar.
It might fit under the yellow-200 if we are not using that one already

hfekete (changes requested):

hfekete (approved):

PR #46, opened 2024-09-19
22059 - loading animation for loading business info.

hfekete:
*Issue:22059 *https://github.com/bcgov/entity/issues/22059 

*Description of changes:*
- loading animation
- test for loading animation


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-46-n5t92760.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-46-n5t92760.web.app

src/app.vue
```diff
@@ -86,11 +88,8 @@ watch(accountErrors.value, (val) => { if (val && val.length > 0) { handleError(v
           <bcros-contact-info class="font-normal font-16 mt-4" :contacts="getContactInfo('registries')" />
         </template>
       </bcros-dialog>
-      <div v-if="appLoading">
-        <UIcon
-          name="i-heroicons-arrow-path"
-          class="animate-spin text-[50px] text-gray-700 absolute top-40 left-[50%]"
-        />
+      <div v-if="dashboardIsLoading || appLoading">
+        <BcrosLoadingIcon class="animate-spin text-6xl text-gray-700 absolute top-40 left-[50%]" />
```
kialj876, line 92: Do you think we could move this to the layout? That way it could include covering the business details header. I think it should work pretty good if we replace the 'appLoading' with your new 'trackUiLoadingStart/Stop' inside the onMounted in here

kialj876 (commented):
Hey nice work, just the one comment if it makes sense to you too

src/stores/ui.ts
```diff
@@ -0,0 +1,23 @@
...
+    uiIsLoading.value.push(trackingUuid)
+  }
+
+  const trackUiLoadingStop = (trackerUuidDone: string) => {
+    const index = uiIsLoading.value.find(tracker => tracker === trackerUuidDone)
+    if (index !== -1) {
+      uiIsLoading.value.splice(index, 1)
+    }
+  }
+
```
patrickpeinanw, line 15: I think we should use `findIndex` since `find` will return the value not the index.

cypress/e2e/initial-page-load.cy.ts
```diff
@@ -1,12 +1,38 @@
...
+    cy.interceptAddresses(legalType).as('getAddresses')
+    cy.interceptParties(legalType, isHistorical).as('getParties')
+    cy.interceptAffiliationRequests(false, false).as('getAffiliationRequests')
+    cy.interceptTasks('tasksEmpty.json').as('getTasks')
+
+    cy.visit(`/${businessIdentifier}`)
+
+    cy.get('[data-cy="loading-icon"]').should('be.visible').then(() => {
+      cy.wait(['@getBusinessInfo'])
+      cy.get('[data-cy="loading-icon"]').should('not.exist')
```
patrickpeinanw, line 35: question: why do we just wait `@getBusinessInfo`? Do we need to wait for the other calls to be intercepted (like getTasks) before the loading icon disappears?

hfekete: No, mandatory one is to wait for 2nd load business call. Generally we could wait for others too, but the thing is there is ~5 sec wait anyways if it is not immediately available. And if it fails with timeout we should investigate why. So its not necessary here.

src/stores/ui.ts
hfekete: Thank you, this is great spot

src/app.vue
hfekete: moved them to the layouts, and replaced main loading with track loading stuff.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-46-n5t92760.web.app

patrickpeinanw (approved):
Nice work.

PR #54, opened 2024-10-22
change address and director handling

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/21889

*Description of changes:*
Added a carded (UCard) dialog option, added dialog that shows on change address. Added redirection rules for addresses and director changes in all cases. NOTE this does not implement a page for those in this dashboard those pages still only exist in the current/older dashboard

### DIFF
--- a/src/components/bcros/dialog/CardedModal.vue
+++ b/src/components/bcros/dialog/CardedModal.vue
@@ -1,0 +1,79 @@
+<template>
+  <UModal
+    :attach="attach || ''"
+    :model-value="display"
+    :data-cy="'bcros-dialog' + (name ? `-${name}` : '')"
+  >
+    <UCard
+      v-if="options"
+      :ui="{
+        header: {
+          background: 'bg-bcGovColor-darkBlue'
+        }
+      }"
+    >
+      <template #header>
+        <UIcon v-if="options.alertIcon" name="i-mdi-information-outline" class="text-4xl text-red-500 mb-2" />
+        <h2 data-cy="bcros-dialog-title" class="text-white">
+          {{ options.title }}
+        </h2>
+        <UButton
+          v-if="!options.hideClose"
+          color="primary"
+          class="absolute top-0 right-0"
+          icon="i-heroicons-x-mark-20-solid"
+          variant="ghost"
+          data-cy="bcros-dialog-close-btn"
+          @click="close()"
+        />
+      </template>
+
+      <div data-cy="bcros-dialog-text">
+        <!-- can be replaced with <template v-slot:content> -->
+        <slot name="content" :options="options">
+          <dialog-content
+            :base-text="options.text"
+            :extra-text="options.textExtra"
+          />
+        </slot>
+      </div>
+      <slot name="extra-content" :options="options" />
+
+      <template #footer>
+        <!-- can be replaced with <template v-slot:buttons> -->
+        <slot name="buttons" :options="options">
+          <div class="flex justify-center gap-5">
+            <div v-for="button, i in options.buttons" :key="'dialog-btn-' + i">
+              <slot :name="'dialog-btn-slot-' + button.slotId">
+                <dialog-button :button="button" data-cy="bcros-dialog-btn" @close="emit('close')" />
+              </slot>
+            </div>
+          </div>
+        </slot>
+      </template>
+    </UCard>
+  </UModal>
+</template>
+
+<script setup lang="ts">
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-54-dvla6j1b.web.app

src/components/bcros/dialog/CardedModal.vue
```diff
@@ -0,0 +1,79 @@
...
+      v-if="options"
+      :ui="{
+        header: {
+          background: 'bg-bcGovColor-darkBlue'
+        }
+      }"
+    >
+      <template #header>
+        <UIcon v-if="options.alertIcon" name="i-mdi-information-outline" class="text-4xl text-red-500 mb-2" />
+        <h2 data-cy="bcros-dialog-title" class="text-white">
```
hfekete, line 17: I would wrap `h2` into the `<slot>` so next person could reuse/change that with template if needed

src/pages/dashboard.vue
```diff
@@ -174,10 +174,94 @@ const pendingAddress = computed(() => {
 
 const isChangeAddressDisabled = computed(() => business.currentBusiness.adminFreeze || pendingAddress.value)
 const isChangeDirectorDisabled = computed(() => business.currentBusiness.adminFreeze)
+const showChangeOfAddress = ref(false)
+
+const toggleChangeOfAddress = (show: boolean) => {
```
hfekete, line 179: just a note for future, toggle usually switches states if it was off -> on; false -> true.
Here, name would be maybe bit more self documenting if you would name method `isShowChangeOfAddress`

src/pages/dashboard.vue
```diff
@@ -174,10 +174,94 @@ const pendingAddress = computed(() => {
...
+  navigateTo(url, { external: true })
+}
+
+const changeAddress = () => {
+  if (business.isEntityFirm()) {
+    const baseUrl = useRuntimeConfig().public.editApiURL
+    const url = `${baseUrl}${business.currentBusinessIdentifier}/change`
+    navigateTo(url, { external: true })
+  } else if (business.isBaseCompany()) {
+    toggleChangeOfAddress(true)
```
hfekete, line 195: e.g. here, when I read, its like toggleChangeOfAddress, param true meaning it will toggle it.

src/components/bcros/dialog/CardedModal.vue
hfekete: another reason for wrapping it in slot is, for example if I would pass background white for the header, I would not be able to change text color of h2.

hfekete (approved):
Good work, had few comments/notes.

hfekete:
Uuuh, I forgot to mention, please add at least one cypress test for displaying/hiding the modal.

src/components/bcros/dialog/CardedModal.vue
```diff
@@ -0,0 +1,79 @@
...
+    :attach="attach || ''"
+    :model-value="display"
+    :data-cy="'bcros-dialog' + (name ? `-${name}` : '')"
+  >
+    <UCard
+      v-if="options"
+      :ui="{
+        header: {
+          background: 'bg-bcGovColor-darkBlue'
+        }
```
patrickpeinanw, line 12: The header needs a rounded top corners.  This is what we have for the freezeUnfreeze modal: `:ui="{ header: { background: 'bg-bcGovColor-darkBlue', base: 'font-2xl font-bold text-white rounded-t-lg' } }`.   Actually, I think we should use this CardedModal in the freezeUnfreeze component, but we can worry about this later.

patrickpeinanw (approved):
Looks good to me. Just one minor comment on the modal style. Also, we've been using the `<BcrosDialog>` for all the pop-up modals in the new dashboard (except for the FreezeUnfreeze modal). We should discuss with the PO and UI/UX later to decide which one we want to use.

src/components/bcros/dialog/CardedModal.vue
BrandonSharratt: Good idea, as it stands you can't colour the header so I made that changeable too

src/pages/dashboard.vue
BrandonSharratt: is to met would return the value so I've changed it to be setChangeOfAddress

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-54-dvla6j1b.web.app

PR #60, opened 2024-10-28
fix dialog title, remove red border from correction todo and disable …

BrandonSharratt:
…change when a correction filing is in todo

*Issue:*https://github.com/bcgov/entity/issues/23948
https://github.com/bcgov/entity/issues/23472

*Description of changes:*
Removed red border error style from Correction to do items, disabled change address / director buttons when a correction draft/pending exists. fixed dialog title for dissolution

### DIFF
--- a/src/pages/dashboard.vue
+++ b/src/pages/dashboard.vue
@@ -188,21 +188,24 @@ const pendingAddress = computed(() => {
 
 const hasProgressCorrections = computed(() => {
   if (filings && filings.value && filings.value.length > 0) {
-    filings.value.filter((filing) => {
+    console.log('hi1')
+    const corrections = filings.value.filter((filing) => {
       return filing.name === FilingTypes.CORRECTION &&
         (filing.status === FilingStatusE.DRAFT ||
         filing.status === FilingStatusE.PENDING ||
         filing.status === FilingStatusE.PENDING_CORRECTION)
     })
+    return corrections.length > 0
   }
+  console.log('hi2')
   return false
 })
 
-const isChangeAddressDisabled = computed(() => {
-  return business.currentBusiness.adminFreeze || pendingAddress.value || hasProgressCorrections
-})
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-60-9h22rajc.web.app

src/pages/dashboard.vue
```diff
@@ -186,8 +186,24 @@ const pendingAddress = computed(() => {
...
+    return corrections.length > 0
+  }
+  return false
+})
+
+const isChangeAddressDisabled = computed(() =>
+  business.currentBusiness.adminFreeze || pendingAddress.value || hasProgressCorrections.value
+)
+
+const isChangeDirectorDisabled = computed(() => business.currentBusiness.adminFreeze || hasProgressCorrections.value)
```
patrickpeinanw, line 206: I think this ticket is only for the red border in the todo section. The 'Change' button in the address section will be handled in #24010. The screenshot shows all the mismatches, but we broke the original QA bug into different ticket. 

Also, the logic for `isChangeAddressDisabled` and `isChangeDirectorDisabled` seems wrong. 

https://github.com/bcgov/business-filings-ui/blob/3a62b2b470bf16c181b879057823ff795e5faf68/src/views/Dashboard.vue#L131-L145  it uses: `:disabled="!isAllowed(AllowableActions.ADDRESS_CHANGE)"`. The function `isAllowed()` is already implemented in the business store.

BrandonSharratt: The ticket I grabbed mentioned the change buttons still so I added handling for the case when a correction filing exists. (It felt large for a 1) The logic in isChangeAddressDisabled and isChangeDirectorDisabled was already there I just added the hasProgressCorrections flag

BrandonSharratt: We will handle the other cases in #24010

patrickpeinanw (approved):

PR #66, opened 2024-11-15
24293 - verify auth access to entity

hfekete:
*Issue:24293 *https://github.com/bcgov/entity/issues/24293 

*Description of changes:*
- verify access to entity
- redirect to error page (to stop loading other elements)
- show error modal
- update tests


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-66-tbrt2fr3.web.app

src/components/bcros/dialog/CardedModal.vue
```diff
@@ -15,19 +15,21 @@
...
-          <UButton
-            v-if="!options.hideClose"
-            color="primary"
-            class="absolute top-0 right-0"
-            icon="i-heroicons-x-mark-20-solid"
-            variant="ghost"
-            data-cy="bcros-dialog-close-btn"
-            @click="close()"
-          />
+          <slot name="header">
```
patrickpeinanw, line 18: Why there is an additional slot inside the original header slot?

src/stores/account.ts
```diff
@@ -21,6 +22,22 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
...
   const apiURL = useRuntimeConfig().public.authApiURL
 
+  async function verifyAccountAuthorizations (identifier?: string): Promise<boolean> {
+    if (!identifier) {
+      accountErrors.value.push(AccountAccessError)
+      return false
+    }
+
+    const x = await useBcrosFetch(`${apiURL}/entities/${identifier}/authorizations`, {})
+      .then((response) => {
```
patrickpeinanw, line 32: I would use a more descriptive variable name than 'x'.

src/interfaces/error-i.ts
```diff
@@ -8,3 +8,10 @@ export interface ErrorI {
   statusCode: number | null,
   type?: ErrorCodeE
 }
+
+export const AccountAccessError: ErrorI = {
```
patrickpeinanw, line 12: shall we move this const outside of the interface definition? like to the account store.

patrickpeinanw (commented):
Nice work. I just have some comments and questions.

src/interfaces/error-i.ts
hfekete: I would keep this here unless you have specific reason to move it out ?

src/components/bcros/dialog/CardedModal.vue
hfekete: Nice catch, I think this was from merge when I merged my local branches :P

cypress/e2e/components/filings/addStaffFiling.cy.ts
```diff
@@ -142,27 +145,31 @@ context('Add Staff Filing', () => {
         response).as('businessComments')
     })
     cy.visitBusinessDashFor('businessInfo/ben/active.json', undefined, false, false, undefined, allFilings, true)
-    cy.get('[data-cy="add-staff-filing"]').should('exist').click()
+    cy.get('[data-cy="add-staff-filing"]').should('exist')
+    cy.get('[data-cy="add-staff-filing"]').click()
```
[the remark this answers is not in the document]
hfekete: In some places, granted, this one might not.
"should" should be safe to chain on, but for example chaining click().get etc ... is not safe. Though it could work, depending on how dom behaves at that moment. 

I just tried to clean it up for reading comprehension a bit, so I could understand better whats happening. 

What is your preferred way of reading it, do you find it more easily to read like this, or when the commands are chained one after the other ?

patrickpeinanw (approved):
Thank for all the changes!

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-66-tbrt2fr3.web.app

PR #78, opened 2024-11-25
fix not in good standing staff change business information

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/24459

*Description of changes:*
Changed the function to not allow staff to change business info on not in good standing businesses

### DIFF
--- a/src/components/bcros/businessDetails/Links.vue
+++ b/src/components/bcros/businessDetails/Links.vue
@@ -35,9 +35,7 @@ const isPendingDissolution = computed(() => {
 })
 
 const isChangeBusinessInfoDisabled = computed(() => {
-  if (!currentBusiness.value.goodStanding) {
-    // todo: add staff exclusion. Staff should not be allowed to skip rules for business
-    // as if this is enabled, and not in good standing, only thing that will come is popup warning
+  if (!currentBusiness.value.goodStanding && !hasRoleStaff) {
     return false
   }
 


hfekete (approved):

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-78-cn2he8w8.web.app

src/components/bcros/businessDetails/Links.vue
```diff
@@ -35,9 +35,7 @@ const isPendingDissolution = computed(() => {
 })
 
 const isChangeBusinessInfoDisabled = computed(() => {
-  if (!currentBusiness.value.goodStanding) {
-    // todo: add staff exclusion. Staff should not be allowed to skip rules for business
-    // as if this is enabled, and not in good standing, only thing that will come is popup warning
+  if (!currentBusiness.value.goodStanding && !hasRoleStaff) {
```
severinbeauvais, line 38: I'm confused by the logic: "if the business is not in good standing and this isn't staff then don't disable the action" ?

Also, please compare your temporary URL with the dev one -- specifically, click on View and Change Business Information. Why do I see the voluntary dissolution dialog?

Note that, in Dev, View and Change Business Information is enabled when I am staff - see https://dev.business.bcregistry.gov.bc.ca/BC0870818/.

BrandonSharratt: This logic is pulled right out of business-filings-ui https://github.com/bcgov/business-filings-ui/blob/29cc2f6b204f3755ad13eafd45aa4e524337b071/src/components/EntityInfo/EntityMenu.vue#L400

The example you sent is allowed because the staff user falls into the base case return !this.isAllowed(AllowableActions.BUSINESS_INFORMATION)

PR #80, opened 2024-11-26
'Consent to Continue Out' - fix filing history text and display badge in tombstone

patrickpeinanw:
*Issue:*https://github.com/bcgov/entity/issues/24375

*Description of changes:*
- Fix the text display for Consent to Continue Out filing
- Display "AUTHORIZED TO CONTINUE OUT" in the tombstone

Example business: BC0883688
image

### DIFF
--- a/src/stores/filings.ts
+++ b/src/stores/filings.ts
@@ -14,6 +14,24 @@ export const useBcrosFilings = defineStore('bcros/filings', () => {
   const apiURL = useRuntimeConfig().public.legalApiURL
 
   const downloadingInProgress = ref(false)
+
+  /** Whether the business is authorized to continue out, i.e. true if cco expiry date is present or in the future. */
+  const isAuthorizedToContinueOut = computed(() => {
+    const ccoFiling = filings.value?.find((val) => {
+      const exp = val.data?.consentContinuationOut?.expiry
+      if (exp) {
+        return true
+      }
+      return false
+    })
+    if (ccoFiling) {
+      const exp = ccoFiling.data?.consentContinuationOut?.expiry
+      const ccoExpiryDate = apiToDate(exp)
+      return ccoExpiryDate >= new Date()
+    }
+    return false
+  })
+
   const setDownloadingInProgress = (isDownloading: boolean) => {
     downloadingInProgress.value = isDownloading
   }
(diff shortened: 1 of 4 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-80-h3y8h0dp.web.app

src/stores/filings.ts
```diff
@@ -14,6 +14,24 @@ export const useBcrosFilings = defineStore('bcros/filings', () => {
...
+      const exp = val.data?.consentContinuationOut?.expiry
+      if (exp) {
+        return true
+      }
+      return false
+    })
+    if (ccoFiling) {
+      const exp = ccoFiling.data?.consentContinuationOut?.expiry
+      const ccoExpiryDate = apiToDate(exp)
+      return ccoExpiryDate >= new Date()
```
severinbeauvais, line 30: You have to be careful when you use `new Date()` because it gets the date-time from the user's web browser, which could be wrong (accidentally or intentionally). You may have seen some code in other Entity UIs where the date-time is fetched from the Legal API, to mitigate this (but I don't really like that design).

The risk here is that you'd allow a continuation out after the expiry date-time, but hopefully the Legal API will validate the filing and reject it accordingly.

If you have any ideas on how to compare "known good" dates, raise it up :)

severinbeauvais (approved):

hfekete (approved):

src/stores/filings.ts
hfekete: I am with you on this one. Dates are hard, and extra care is needed to make them good.

I was thinking of rewriting date util functions using Luxon. As for any date activity, the best approach usually is to use library that has faced and solved this issues.

NB: It looks like there are few new libraries that are overtaking Luxon and Moment (which I used before) as go to libraries for dates. It would probably be good that some investigation is done if those libraries are on par and support complex timezone stuff. (some new kids on the block: date-fns, dayjs, etc ...)

severinbeauvais: I did not use Moment in the Entity UIs because it's not longer needed -- JS can now handle dates just fine (though not always straighforward). And Luxon is just an evolution.

Even with these libraries, the root cause is not handled: we should not trust the browser's date-time. That's why some Entity UIs poll (using HEAD call) to get the date-time from the API, but it feels klutzy. There must be a library that does this for us in the background (NTP for UI???).

For now I guess we're OK (because API validates the filing), but if you find anything, let's discuss it!

PR #94, opened 2024-12-04
Add full loader for download business summary

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/24518

*Description of changes:*
Added full screen modal and made it display when downloading business summary and hide when finished. Note that while there does seem to be a downloading flag for the common template, it doesn't appear to be passed through properly so I can't simply set it to true to cause them all to download and there is quite the chain that I'd have to add it to props and pass it through on so I went this way as it's easier

### DIFF
--- a/src/components/bcros/filing/addStaffFiling/Index.vue
+++ b/src/components/bcros/filing/addStaffFiling/Index.vue
@@ -149,7 +149,7 @@ const allActions: ComputedRef<Array<MenuActionItem>> = computed(() => {
       datacy: 'continue-out',
       label: t('label.filing.staffFilingOptions.continueOut'),
       click: () => {
-        goToBusinessDashboard(`/${currentBusiness.value.identifier}/consent-continuation-out`, { filingId: '0' })
+        goToBusinessDashboard(`/${currentBusiness.value.identifier}/continuation-out`, { filingId: '0' })
       }
     },
     { // <!-- Extend Limited Restoration  -->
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-94-ziv1i9va.web.app

src/components/bcros/businessDetails/Links.vue
```diff
@@ -243,7 +262,7 @@ const contacts = getContactInfo('registries')
     </BcrosDialog>
 
     <!-- Staff Comments -->
-    <div v-if="hasRoleStaff && currentBusiness && !isDisableNonBenCorps()">
```
severinbeauvais, line 246: Don't we want to disable staff comments for non-Ben corps (that should be managed in COLIN instead)?

severinbeauvais:
While testing this, I found a new bug. Try this business (which is actually in Test only). Compare with Filings UI.

https://business-dashboard-dev--pr-94-ziv1i9va.web.app/BC1151476

src/components/bcros/businessDetails/Links.vue
severinbeauvais: It's not enabled in Filings UI. LMK if this is a change in requirements.

image

BrandonSharratt: Accidently chucked this commit when adding a new one, but I didn't change that line

PR #110, opened 2024-12-11
24458 - no expansion panel for pending correction filing in the todo section

patrickpeinanw:
*Issue:*https://github.com/bcgov/entity/issues/24458

*Description of changes:*
When there is a pending correction in the todo section, no expansion panel should be displayed and there should be no 'view detail' button. 

It is hard to reproduce because a correction filing is added to the filing history right after the filing is submitted. 
Cypress test is used to mock this scenario. See screenshot below:
pending correction

### DIFF
--- a/src/utils/todo/task-filing/content-loader.ts
+++ b/src/utils/todo/task-filing/content-loader.ts
@@ -183,8 +183,7 @@ export const addExpansionContent = (todoItem: TodoItemI): void => {
     // if it is a draft correction filing
     todoItem.expansionContent = TodoExpansionContentE.DRAFT_CORRECTION
   } else if (todoItem.name === FilingTypes.CORRECTION) {
-    // if it is a correction filing (non-draft)
-    todoItem.expansionContent = TodoExpansionContentE.CORRECTION
+    // if it is a non-draft correction filing (pending filing), no expansion panel in this case
   } else if (
     (todoItem.status === FilingStatusE.DRAFT || todoItem.status === FilingStatusE.APPROVED) && todoItem.nameRequest
   ) {
--- a/cypress/e2e/components/todos/pending.cy.ts
+++ b/cypress/e2e/components/todos/pending.cy.ts
@@ -174,4 +174,22 @@ context('TODOs -> Pending Filing', () => {
       .eq(0).click()
       .wait('@cancelPayment')
   })
+
+  it('Paid and pending correction', () => {
+    cy.visitBusinessDashFor('businessInfo/ben/active.json', undefined, false, false, 'pendingCorrection.json')
+
+    cy.get('[data-cy="header_todo"]').should('exist')
+    cy.get('[data-cy="todoItemList"]').should('exist')
+
+    // subtitle
+    cy.get('[data-cy^="todoItem-label-"]')
+      .should('exist')
+      .should('contains.text', 'FILING PENDING')
+
+    // no 'View Details' button
+    cy.get('[data-cy^="todoItem-showMore-"]').should('not.exist')
+
+    // no action button
+    cy.get('[data-cy^="todoItemActions-"]').find('button').should('not.exist')
+  })
 })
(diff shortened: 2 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-110-5hxdkgcd.web.app

src/utils/todo/task-filing/content-loader.ts
```diff
@@ -183,8 +183,7 @@ export const addExpansionContent = (todoItem: TodoItemI): void => {
     // if it is a draft correction filing
     todoItem.expansionContent = TodoExpansionContentE.DRAFT_CORRECTION
   } else if (todoItem.name === FilingTypes.CORRECTION) {
-    // if it is a correction filing (non-draft)
-    todoItem.expansionContent = TodoExpansionContentE.CORRECTION
+    // if it is a non-draft correction filing (pending filing), no expansion panel in this case
   } else if (
```
patrickpeinanw, line 187: this `addExpansionContent` function determines what expansion panel to be rendered for the todo item. It covers the logic in https://github.com/bcgov/business-filings-ui/blob/f9747e4b20d50da231dcd8a45d1feb45ffa5a498/src/components/Dashboard/TodoList.vue#L589-L691   In TodoList.vue L623-L634, there is a template for a correction todo item that is not draft. However, this expansion panel is never rendered in the Filing UI. In TodoList.vue of Filing UI, the 'View Detail' button is controlled by `showDetailsBtnRed()` (L929) and `showDetailsBtnBlue()` (L900).

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-110-eurm1gxn.web.app

src/utils/todo/task-filing/content-loader.ts
severinbeauvais: The code block you reference in TodoList.vue is legacy: prior to handling corrections via a regular filing (in Edit UI) and the corresponding back end code, we supported only limited corrections as "local filings" (in Filings UI) that only accepted a detail comment, and Ops Team would need to manually make the changes in the db. This code block handled the case of a correction filing not in draft, ie, staff "filed" it and now Ops Team has to action it (and then change this correction task to a correction ledger item).

Ideally, someone would test the various states of a correction filing (draft, pending, paid, etc) and ensure that it displays correctly in the dashboard.

(Note that a regular user cannot action a correction task; only staff can. I didn't test that in the new dashboard.)

patrickpeinanw: @severinbeauvais We have this business in TEST (https://test.business-dashboard.bcregistry.gov.bc.ca/BC1051272) that has a pending correction filing in Todo. You can see it has the 'View Details' button which is not supposed to appear since there is no expansion panel in this case. -- this ticket is to solve this issue.

Do you know a way to create a pending correction just like what we see in the TEST example?

severinbeauvais: Re:
image

There is no way to create a new pending Correction in the To Do list using current flows. This is a left-over filing from a couple of years ago, before we fully implemented correction filing processing.

I suppose you could create such a filing manually in the Legal db. (Eg, start from a normal correction task and update its state.)

Again, there shouldn't be any Correction tasks in this state since a couple of years ago. It would be wise to confirm this in the Prod Lear db (but I don't have access to it).

Given this, I guess your code change is "safe" but won't be used, which is ok.

severinbeauvais: Since correction filing processing is now implemented, these old pending corrections should be deleted from all dbs.

patrickpeinanw: @severinbeauvais Thanks for your clarification.

hfekete (approved):
Good job for using cypress tests to create situations that are hard to achieve.

PR #115, opened 2025-01-01
24451 - Add the 'Fetching Data' spinner for downloading business summary and loading document lists

patrickpeinanw:
*Issue:*https://github.com/bcgov/entity/issues/24451

*Description of changes:*
Re-organize code for the loading state spinner.
- Show "Loading Dashboard" spinner when the page is loading
- Show "Fetching Data" when the 1) downloading a business summary and 2) loading a document list

### DIFF
--- a/src/components/bcros/businessDetails/Links.vue
+++ b/src/components/bcros/businessDetails/Links.vue
@@ -22,8 +22,8 @@ const { isAllowedToFile, isDisableNonBenCorps } = useBcrosBusiness()
 const isCommentOpen = ref(false)
 const isDissolutionDialogOpen = ref(false)
 const { goToCreatePage } = useBcrosNavigate()
+const ui = useBcrosDashboardUi()
 const filings = useBcrosFilings()
-const isFetchingDataSpinner = ref(false)
 
 const isAllowedBusinessSummary = computed(() =>
   !!currentBusinessIdentifier.value &&
@@ -122,7 +122,7 @@ const promptChangeBusinessInfo = () => {
 
 /** Request and Download Business Summary Document. */
 const downloadBusinessSummary = async (): Promise<void> => {
-  isFetchingDataSpinner.value = true
+  ui.fetchingData = true
   const businessId = currentBusiness.value.identifier
   const apiURL = useRuntimeConfig().public.legalApiURL
   const summaryDocument: DocumentI = {
@@ -135,7 +135,7 @@ const downloadBusinessSummary = async (): Promise<void> => {
   if (blob) {
     saveBlob(blob, summaryDocument.filename)
   }
-  isFetchingDataSpinner.value = false
+  ui.fetchingData = false
 }
 
 /** Creates a draft filing and navigates to the Create UI to file a company dissolution filing. */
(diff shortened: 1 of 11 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-115-d5ytvfb9.web.app

src/components/bcros/businessDetails/Links.vue
```diff
@@ -135,7 +135,7 @@ const downloadBusinessSummary = async (): Promise<void> => {
   if (blob) {
     saveBlob(blob, summaryDocument.filename)
   }
-  isFetchingDataSpinner.value = false
+  ui.fetchingData = false
```
severinbeauvais, line 138: If line 134 throws an error, this line won't be executed and the user will be stuck.

I suggest either a try-catch-finally block, or something like this:
`const blob = await fetchDocuments(summaryDocument.link).catch(() => null)`

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -109,9 +110,16 @@ const isStatusApproved = computed(() => isFilingStatus(filing.value, FilingStatu
...
-const showDetails = () => {
+const showDetails = async () => {
   if (filing.value.documents === undefined && filing.value.documentsLink) {
-    loadDocumentList(filing.value)
+    ui.fetchingData = true
+
+    await loadDocumentList(filing.value)
+
+    // wait for another 500ms to show the loading modal
+    await new Promise(resolve => setTimeout(resolve, 250))
```
severinbeauvais, line 120: I recommend moving this to a common util file, and name it "sleep" to be consistent with other UIs.

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -109,9 +110,16 @@ const isStatusApproved = computed(() => isFilingStatus(filing.value, FilingStatu
...
+
+    ui.fetchingData = false
```
severinbeauvais, line 122: As above, if line 117 throws an exception, this spinner will never be disabled.

src/components/bcros/filing/common/HeaderActions.vue
```diff
@@ -323,14 +326,21 @@ const actions: any[][] = [[
...
   if (filing.value.documents === undefined && filing.value.documentsLink) {
-    loadDocumentList(filing.value)
+    ui.fetchingData = true
+
+    await loadDocumentList(filing.value)
+
+    // wait for another 500ms to show the loading modal
+    await new Promise(resolve => setTimeout(resolve, 500))
+
+    ui.fetchingData = false
```
severinbeauvais, line 343: Here, too.

src/layouts/default.vue
severinbeauvais, line 1: Is this a duplicate of business.vue, above? Do we need a duplicate?

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -109,9 +110,16 @@ const isStatusApproved = computed(() => isFilingStatus(filing.value, FilingStatu
...
 
-const showDetails = () => {
+const showDetails = async () => {
   if (filing.value.documents === undefined && filing.value.documentsLink) {
-    loadDocumentList(filing.value)
+    ui.fetchingData = true
+
+    await loadDocumentList(filing.value)
+
+    // wait for another 500ms to show the loading modal
```
severinbeauvais, line 119: This looks like a port of the code from Filings UI. I'd like to see a better explanation of why you have the "sleep" here, ie:

image

src/layouts/default.vue
patrickpeinanw: Currently, the dashboard page uses the business.vue, and the error page uses default.vue. 

default.vue should only contain the basic elements like header, breadcrumb, and footer. In my recent commit, I removed the dialog components in default.vue. But you are right, default.vue and business.vue are still very similar. @hfekete what is your thoughts on this? should we just use one layout template?

src/components/bcros/businessDetails/Links.vue
patrickpeinanw: Thanks for catching this. Error-handling has been added now. I have created a new ticket #25125 for a downloading error dialog.

severinbeauvais (approved):

src/utils/sleeps.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * "Sleeps" for specified timeout. Must be awaited.
+ * @param ms Delay to sleep, in milliseconds.
+ * @returns A promise to await upon.
+ */
+export function sleep (ms: number): Promise<void> {
```
hfekete, line 6: just a note for the future, things like this, you can put in composable, then they are autoloaded and you can just use them as any inbuilt function !

hfekete (approved):
LGTM, nice work.

PR #117, opened 2025-01-10
Business Dashboard UI: implement background retry to check for latest state

Rajandeep98:
*Issue:*https://github.com/bcgov/entity/issues/23968

*Description of changes:*

every 1 second until the 10 second mark
every 10 seconds until the 1 minute mark
every minute until the 30 minute mark
every hour thereafter
data polling slim end point , compare it with initial /identifier api endpoint's modified date, 
show toast msg if modified date vs slim endpoint modified date > 0s

### DIFF
--- a/src/app.config.ts
+++ b/src/app.config.ts
@@ -88,21 +88,27 @@ export default defineAppConfig({
       base: 'xs:min-w-[90vw] md:min-w-[720px] text-gray-700'
     },
     notification: {
-      title: 'text-white',
+      title: 'text-sm font-medium text-white max-w-full',
       description: 'text-white',
+      actions: 'flex items-center gap-2 mt-3 flex-shrink-0',
       background: 'bg-gray-700',
+      rounded: 'rounded',
+      ring: '',
       progress: {
         background: 'bg-transparent'
       },
       default: {
         closeButton: {
-          class: 'hover:text-gray-100',
-          color: 'gray'
+          icon: null,
+          label: 'Close',
+          variant: 'outline'
         }
       }
     },
     notifications: {
-      position: 'bottom-5 left-[40%]'
+      position: 'top-20 bottom-[unset]',
+      width: 'w-full sm:w-5/12',
+      container: 'px-4 sm:px-6 py-6 space-y-3 overflow-y-auto'
     },
     radio: {
       base: 'h-5 w-5 mt-[3px]',
@@ -208,7 +214,45 @@ export default defineAppConfig({
     },
     button: {
       base: 'focus:outline-none focus-visible:outline-0 disabled:cursor-not-allowed ' +
-        'disabled:opacity-35 aria-disabled:cursor-not-allowed aria-disabled:opacity-75 flex-shrink-0'
+        'disabled:opacity-35 aria-disabled:cursor-not-allowed aria-disabled:opacity-75 flex-shrink-0',
+      variant: {
+        outline: `hover:bg-text-white-50 
+          disabled:bg-transparent 
+          dark:hover:bg-text-white-950 
+          dark:disabled:bg-transparent 
+          font-sm shadow-sm 
+          ring-1 ring-inset ring-gray-300 
+          dark:ring-gray-700 text-xs font-medium rounded 
+          text-white tracking-wide py-1 px-2 
+          disabled:text-text-white-500 
+          dark:text-white 
+          dark:hover:text-bcGovBlue-300 
+          dark:disabled:text-text-white-400 
+          focus-visible:ring-2 
+          focus-visible:ring-inset 
+          focus-visible:ring-text-white-500 
+          dark:focus-visible:ring-white 
+          inline-flex items-center`,
+
+        refresh: `bg-white hover:bg-text-blue-500 
+          disabled:bg-transparent 
+          dark:hover:bg-text-blue-500 
+          dark:disabled:bg-transparent 
+          font-sm shadow-sm 
+          ring-1 ring-inset ring-blue-500 
+          dark:ring-blue-500 text-xs font-medium rounded 
+          text-blue-500 tracking-wide py-1 px-2 
+          disabled:text-blue-500 
+          dark:text-blue-500 
+          dark:hover:text-bcGovBlue-300 
+          dark:disabled:text-text-white-400 
+          focus-visible:ring-2 
+          focus-visible:ring-inset 
+          focus-visible:ring-text-white-500 
+          dark:focus-visible:ring-white 
+          inline-flex items-center px-2 py-1.5`
+
+      }
     }
   }
 })
--- a/src/app.config.ts
+++ b/src/app.config.ts
@@ -101,7 +101,7 @@ export default defineAppConfig({
         closeButton: {
           icon: null,
           label: 'Close',
-          variant: 'outline'
+          variant: 'cancel'
         }
       }
     },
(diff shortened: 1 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

src/stores/business.ts
```diff
@@ -67,9 +67,10 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
   }
 
   /** Return the business details for the given identifier */
-  async function getBusinessDetails (identifier: string, params?: object) {
+  async function getBusinessDetails (identifier: string, params?: object, slim: boolean = false) {
+    const url = `${apiURL}/businesses/${identifier}${slim ? '?slim=true' : ''}`
```
severinbeauvais, line 71: Is this function the one use to load all business data? Won't it break other functionality if you're now fetch the slim results instead of full results?

src/pages/dashboard.vue
```diff
@@ -79,6 +82,66 @@ const containRole = (roleType) => {
...
+}
+
+const fetchBusinessDetailsWithDelay = async (identifier: string) => {
+  // Fetch business details and update finalDateString
+  const slimBusiness = await business.getBusinessDetails(identifier, undefined, true)
+  const finalDate = new Date(slimBusiness.lastModified)
+  finalDateString.value = finalDate.toISOString().split('.')[0]
+
+  if (computeTimeDifference(business.initialDateString, finalDateString.value) > 0) {
+    toast.add({
```
severinbeauvais, line 100: If the user doesn't acknowledge the first toast, do more toasts get added on top of it?

severinbeauvais:
Please update the app version before merging.

severinbeauvais (commented):
~Something isn't working right. I tested using the preview link and I don't see the 10 x 1-second polls. I only see the 10-second polls to start.~ _code will be updated_

src/stores/business.ts
severinbeauvais: Answer: see `slim` parameter.

src/app.config.ts
```diff
@@ -208,7 +214,45 @@ export default defineAppConfig({
...
+          dark:text-blue-500 
+          dark:hover:text-bcGovBlue-300 
+          dark:disabled:text-text-white-400 
+          focus-visible:ring-2 
+          focus-visible:ring-inset 
+          focus-visible:ring-text-white-500 
+          dark:focus-visible:ring-white 
+          inline-flex items-center px-2 py-1.5`
+
+      }
```
severinbeauvais, line 255: That's a ton of CSS...

**BTR guys, is there a simpler way to do this?**

src/pages/dashboard.vue
```diff
@@ -116,25 +116,30 @@ const fetchBusinessDetailsWithDelay = async (identifier: string) => {
...
+  const thirdInterval = 60000 // 1 minute
+  const fourthInterval = 3600000 // 1 hour
 
   const poll = () => {
     const elapsedTime = Date.now() - startTime
 
-    if (elapsedTime < 60000) {
-      // Poll every 10 seconds for the first minute
+    if (elapsedTime < 30000) {
+    // Poll every second for the first 30 seconds
```
severinbeauvais, line 128: Do we have agreement that we really need to poll every second for 30 seconds? I thought we said doing this for 10 seconds was enough?

Also, please verify that the PR description matches this logic.

Rajandeep98: resetting back to previous schedule

src/app.config.ts
Rajandeep98: image

patrickpeinanw: The `<UButton>` in Nuxt UI has some default variants https://ui.nuxt.com/components/button#config. In app.config.ts, if you want to modify an existing variant, you can just specify the tailwindcss classes you want to overwrite. If you are adding a new variant (like "refresh"), you can just include the classes that are different from the default values (I guess the default variant for button is "solid"). 

There are multiple places in the codebase where we use variant="outline" for `<UButton>` . Overwriting the styling of the "outline" variant may change the appearance of those buttons. 

Another way to update the styling is using the `:ui` prop. Here is an example https://github.com/bcgov/business-dashboard-ui/blob/35dbb2e17076a062a3facd68428b19ee93a1649d/src/components/bcros/pending/Item.vue#L24-L32  but I am not sure if it is applicable for toast action.

severinbeauvais: @hfekete , @patrickpeinanw Any thoughts about the best way to do this?

src/pages/dashboard.vue
Rajandeep98: every 1 second until the 10 second mark
every 10 seconds until the 1 minute mark
every minute until the 30 minute mark
every hour thereafter

src/app.config.ts
Rajandeep98: added 2 new variants reusing the existing ones, could create issues for the other ui components as these 2 buttons are different than available ones.

src/pages/dashboard.vue
Rajandeep98: it pops up again on next interval

src/pages/dashboard.vue
```diff
@@ -79,6 +82,66 @@ const containRole = (roleType) => {
   )
 }
 
+const computeTimeDifference = (initialDateString, finalDateString) => {
```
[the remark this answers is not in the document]
severinbeauvais: Rajan?

src/pages/dashboard.vue
severinbeauvais: I tested this and there's just the 1 toast (no stacking).

src/stores/business.ts
```diff
@@ -170,21 +171,33 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
     trackUiLoadingStart('businessInfoLoading')
 
     const businessCached = currentBusiness.value && identifier === currentBusinessIdentifier.value
+
     if (!businessCached || force) {
       fetchBusinessComments(identifier)
       currentBusiness.value = await getBusinessDetails(identifier) || {} as BusinessI
+
+      // Converting lastModified values to Date objects
+      const initialDate = new Date(currentBusiness.value.lastModified)
```
severinbeauvais, line 186: Check if there's a Date Mixin or Date Utility (apiToDate) to convert "API datetime" to a Date object.

Directly passing this into a Date object may or may not work in different environments, because the "API datetime" it not exactly in UTC format, ie "2025-01-10T16:31:21.334026+00:00".

src/stores/business.ts
```diff
@@ -170,21 +171,33 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
...
+
+      // truncate milliseconds
+      initialDateString.value = initialDate.toISOString().split('.')[0]
```
severinbeauvais, line 189: Check the Date Mixin to see if you can work with datetimes in their base format.

Or, you could use "slice" to get just the first part of a string.

severinbeauvais (approved):
Still to do:
- check unresolved comment in dashboard.vue
- update app version
- update screenshots

src/app.config.ts
severinbeauvais: Way better. Thanks!

src/stores/business.ts
Rajandeep98: used dateToStringPacific for comparison

src/stores/business.ts
Rajandeep98: updated in comparison keeping the value as it is from api

Rajandeep98:
image

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

src/pages/dashboard.vue
```diff
@@ -79,6 +81,61 @@ const containRole = (roleType) => {
   )
 }
 
+const fetchBusinessDetailsWithDelay = async (identifier: string) => {
+  // Fetch business details and update finalDateString
+  const slimBusiness = await business.getBusinessDetails(identifier, undefined, true)
+  const finalDate = dateToStringPacific(apiToDate(slimBusiness.lastModified))
+  if (dateToStringPacific(apiToDate(business.initialDateString)) !== finalDate) {
```
severinbeauvais, line 88: This is still a string comparison (and may fail because, for example, Dec will come before Jan).

How about comparing Date objects directly? And should you use > instead of ==?

Rajandeep98: going with milliseconds difference after converting to pacific for both time stamps
using getTime()

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

src/pages/dashboard.vue
```diff
@@ -80,6 +82,62 @@ const containRole = (roleType) => {
   )
 }
 
+const fetchBusinessDetailsWithDelay = async (identifier: string) => {
+  // Fetch business details and update finalDateString
+  const slimBusiness = await business.getBusinessDetails(identifier, undefined, true)
```
severinbeauvais, line 87: This looks good enough for now.

Please create a new ticket for the following improvements:
- need to handle possible error from API call above (either with catch statement or catch block)
- should delay first poll by the first poll interval (10s), other we get a regular business call followed immediately by a slim business call, which is unnecessary.

src/pages/dashboard.vue
severinbeauvais: Resolved.

severinbeauvais:
There's a new Cypress test error above. Can you look into it?

cc: @patrickpeinanw , any ideas?

src/pages/dashboard.vue
severinbeauvais: For the second item, could probably just update line 142.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

patrickpeinanw:
> There's a new Cypress test error above. Can you look into it?
> 
> cc: @patrickpeinanw , any ideas?

@Rajandeep98 the error message is "Cannot read properties of null (reading 'lastModified')". In some cases `slimBusiness` or `currentBusiness.value` is null. 

To run and debug cypress tests locally, you first build and run the app (`pnpm run build:local && pnpm run preview`), then do `pnpm run test:e2e`

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

src/pages/dashboard.vue
Rajandeep98: implemented 

- need to handle possible error from API call above (either with catch statement or catch block)

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-117-n0kartyb.web.app

JazzarKarim (approved):
LGTM 👍 

Can you please update the package.json version please?

src/pages/dashboard.vue
severinbeauvais: Resolving!

Rajandeep98:
re-run cypress, all tests passed

severinbeauvais:
@Rajandeep98 , thanks for merging. I forgot to tell you: use the "squash" option when merging so that your PR gets merged as a single commit instead of multiple invididual commits.

PR #127, opened 2025-01-22
25406-Add Enable-Withdrawal-Action Feature Flag

meawong:
*Issue:*https://github.com/bcgov/entity/issues/25406

*Description of changes:*
- Disable "File a Withdrawal" action from drop-down when feature flag is **false**

### DIFF
--- a/src/components/bcros/filing/common/HeaderActions.vue
+++ b/src/components/bcros/filing/common/HeaderActions.vue
@@ -98,7 +98,7 @@ import { z } from 'zod'
 import { type ApiResponseFilingI, FilingStatusE, isFilingStatus, isStaffFiling, isFutureEffective } from '#imports'
 import { FilingCorrectionTypesE } from '~/enums/filing-correction-types-e'
 
-const { getStoredFlag } = useBcrosLaunchdarkly()
+const { getFeatureFlag, getStoredFlag } = useBcrosLaunchdarkly()
 const { hasRoleStaff } = storeToRefs(useBcrosKeycloak())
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-127-wbpxgg9h.web.app

src/components/bcros/filing/common/HeaderActions.vue
```diff
@@ -324,6 +324,11 @@ const goToNoticeOfWithdrawal = () => {
   goToFilingUI(path, params)
 }
 
+const disableWithdrawal = (): boolean => {
+  const ff = getFeatureFlag('enable-withdrawal-action')
```
severinbeauvais, line 328: Why did you use `getFeatureFlag` when line 244 uses `getStoredFlag`?

(I don't actually know when one or the other should be used. @kzdev420 might know.)

JazzarKarim (approved):
LGTM

src/components/bcros/filing/common/HeaderActions.vue
meawong: Kevin recommended using `getFeatureFlag` in this case

severinbeauvais: Weird. That doesn't seem to fit this:

image

Is a change needed?

hfekete: Yes, I believe that it would be better if `getStoredFlag` is used here, I do not think we used `getFeatureFlag` inside the vue files. It basically lets you retrieve flag (if loaded in the store) even if the ldclient is down.

PR #132, opened 2025-01-24
25401 - fix page 500 error if GTM-ID is not valid

JazzarKarim:
*Issue:*https://github.com/bcgov/entity/issues/25401

*Description of changes:*
Changed NUXT_ to VUE_APP
Fix page 500 error in temp url


.env.example
```diff
@@ -36,9 +36,9 @@ VUE_APP_SENTRY_DSN=
 VUE_APP_SENTRY_TRACE_SAMPLE_RATE="1.0"
 
 #google analytics
-NUXT_GTAG_ID=""
-NUXT_GTAG_DEBUG=""
-NUXT_GTM_ID=""
+VUE_APP_GTAG_ID=""
```
JazzarKarim, line 39: Don't know how I missed this. The project is in NUXT because uses VUE_APP for env values. I changed it to stay consistent.

nuxt.config.ts
```diff
@@ -55,14 +55,14 @@ export default defineNuxtConfig({
     ]
   },
   gtm: {
-    enabled: !!process.env.NUXT_GTM_ID?.trim(),
-    id: process.env.NUXT_GTM_ID?.trim() as string,
+    enabled: !!process.env.VUE_APP_GTM_ID?.trim(),
+    id: process.env.VUE_APP_GTM_ID?.trim() || 'GTM-DUMMY', // the dummy value allows app to run if GTM ID could not be loaded
```
JazzarKarim, line 59: If the GTM ID wasn't loaded from 1pass for whatever reason, the project will still load but without analytics. It's OK as we don't want this to block the entire thing from running in case that happens.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-132-acenv075.web.app

nuxt.config.ts
severinbeauvais: Would empty string work?

severinbeauvais:
> https://business-dashboard-dev--pr-132-acenv075.web.app/FM1101299

image

severinbeauvais:
Well, I get to the app now :)

image

severinbeauvais (approved):

.env.example
severinbeauvais: We reviewers missed this, too 😛

nuxt.config.ts
JazzarKarim: Nope, since the error is 
image

The code is expecting a GTM-something.

severinbeauvais: Oh, yeah.

OK, I have no further comment about this one.

PR #133, opened 2025-01-27
25258 Document Record Ledger Link

cameron-eyds:
*Issue:*https://github.com/bcgov/entity/25258

*Notes:*
- This work is feature flagged behind `enable-document-records` 
- Also the nav button to DRS won't render unless the Ledger `filingId` aligns with the `consumerRerenceId` on DRS side.
- Lastly, there will be no links until Entity Corp Filing Submissions are creating Document Records via a DRS hook api side.

*Description of changes:*
- Drs env configuration and setup
- Implements Document Record Btn, IF's, store and requests
- Insertion of Document Record Navigation button into Filing Templates (Ledger) when a Document Record has a consumerReferenceId that matches a filingId. 
- Minor unit testing

Screenshot 2025-01-27 at 10 00 10 AM

Target Destination: Document Record View
Screenshot 2025-01-27 at 10 00 25 AM


.env.example
```diff
@@ -11,12 +11,16 @@ VUE_APP_BCONLINE_URL="https://d1.bconline.gov.bc.ca/"
...
+VUE_APP_DOCUMENTS_UI_URL="https://dev.documents.bcregistry.gov.bc.ca/"
 
 #vaults API
 VUE_APP_AUTH_API_URL="https://auth-api-dev-142173140222.northamerica-northeast1.run.app"
 VUE_APP_AUTH_API_VERSION="/api/v1"
 VUE_APP_PAY_API_URL="https://pay-api-dev.apps.silver.devops.gov.bc.ca"
 VUE_APP_PAY_API_VERSION="/api/v1"
+VUE_APP_DOC_API_URL="https://bcregistry-dev.apigee.net/doc"
+VUE_APP_DOC_API_VERSION="/api/v1"
+VUE_APP_DOC_API_KEY=""
```
severinbeauvais, line 23: Add these to vaults.env file, too.

And ask Patrick to configure the Dev API key in GCP.

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -77,6 +77,16 @@
         </template>
       </slot>
 
+      <slot name="document-record">
+        <!-- if we have document records(aka filings staff completed on behalf of a client), show them -->
+        <template
+          v-if="!isStaffFiling(filing) && enableDocumentRecords && !!getDocIdByFilingId(documents, filing.filingId)"
```
severinbeauvais, line 83: Not staff filing? I think Court Order might be an exception. (It's currently a "simple dialog" staff filing but it's being moved to a one-page filing like the others.)

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -77,6 +77,16 @@
...
+        >
+          <UDivider class="my-6" />
+          <BcrosDocumentRecordBtn :document-id="getDocIdByFilingId(documents, filing.filingId)" />
```
severinbeauvais, line 86: You call this function twice. How about making it a computed?

src/interfaces/document-request-i.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// Define a type for the Axios error response
+export interface ApiErrorIF {
+  message: string
+  status?: number
+  statusText?: string
+  statusCode?: number
+}
+
+export type ApiResponseOrError = ApiResponseIF | ApiErrorIF
```
severinbeauvais, line 51: Nice 👍

src/stores/documents.ts
```diff
@@ -0,0 +1,84 @@
...
+   * Retrieves the consumerDocumentId of a document that matches the given filingId.
+   *
+   * @param documents - An array of Document objects to search through.
+   * @param filingId - The filingId to match against the consumerReferenceId.
+   * @returns The consumerDocumentId of the matching document, or undefined if no match is found.
+  */
+  const getDocIdByFilingId = (documents: DocumentIF[], filingId: number): string | undefined => {
+    const document = documents?.find(doc =>
+      doc.consumerReferenceId === filingId.toString())
+    return document ? document.consumerDocumentId : undefined
```
severinbeauvais, line 75: You could use optional chaining here.

I'm wary of returning undefined, since it may be non-reactive.

.env.example
cameron-eyds: Absolutely, will do!

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: Good question, is Court Order a filing that Staff would do on behalf of a client?
ie a Client has either mailed/emailed/brought in documentation that would than require staff to do on their behalf

src/stores/documents.ts
cameron-eyds: Definitely, will do!

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: Yes.

JazzarKarim (approved):
Aside from Sev's comments, looks good 👍

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: Just to follow up on this, are any other of these Staff Filings also completed at the behest of the client?

Screenshot 2025-01-28 at 9 49 38 AM

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: Updated

src/stores/documents.ts
cameron-eyds: Updated

cameron-eyds:
@severinbeauvais I patched in some correlating filingIds to DRS on Entity: BC0887516, so some of those ledger items should have a document record link out.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: I think Court Order is the only one.

src/stores/documents.ts
severinbeauvais: If possible, please test with `document` or `document.consumerDocumentId` equal to undefined, and check if the computed in CommonTemplate.vue still works correctly.

In the past (albeit in Vue2), I've found that undefined values aren't reactive, meaning getters and watchers don't update as expected.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-133-ww3dxzjc.web.app

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: Ok, thanks for this info sir! 

We currently support Court Order documents on the DRS side i believe, although its not classed as a Corp specific filing (which makes sense). 
Are you comfortable with me collecting a bit more info/requirements from our BA's before enabling on this end?

severinbeauvais: > Are you comfortable with me collecting a bit more info/requirements from our BA's before enabling on this end?

What do you mean, specifically, by "enabling"?

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-133-ww3dxzjc.web.app

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -77,6 +77,14 @@
         </template>
       </slot>
 
+      <slot name="document-record">
+        <!-- if we have document records(aka filings staff completed on behalf of a client), show them -->
+        <template v-if="!isStaffFiling(filing) && enableDocumentRecords && !!getDocumentId">
```
severinbeauvais, line 82: Do you need to check the FF here, or would getDocumentId be falsy without the FF anyway?

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: Oh, you mean just turning on the FF in Dev and maybe Test?

When do you plan to enable this in Prod?

cameron-eyds: I mean by implementing a condition in the logic to allow it for Court Orders, i just want to follow up on scope, requirements and support on our end.

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: I think the latter is true yes, but the computed value also considers that the current user is staff.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: OK, but be advised we're currently working on the stand-alone Court Order filing. It will be just like the other stand-alone filings in Filings UI (except staff-only).

Thanks for the follow-up.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: Which computed value considers whether the current user is staff?

I'm simply thinking you don't need to check the FF here.

I just really paid attention to "isStaffFiling" and I'm thinking it doesn't do what you want. If I'm not mistaken it returns True for Court Order, Registrar's Notation, etc. What you really want here isn't the type of filing, but who filed it, right?

Maybe you don't need to check "isStaffFiling" either, here, because the "simple staff filings" won't have a DRS section, while fuller filings will (when completed by staff).

image

I'm OK to take this offline if you want to chat directly about it.

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: Thanks sir, i will follow up on this. 
Saving these docs to the DRS vs Minio will happen eventually but you raise a good question.

 If this filing is done by staff due to a received paper form which they have need to apply a barcode to before scanning, it would most likely require a Document Id capture component. 
Will follow up with the BA's 100%!

severinbeauvais: Zactly.

What about the other questions here:
- do you need the FF check or is it redundant?
- do you need the staff filing check or is it redundant?

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: We will want to check if the current account is Staff. The DRS UI isn't directly exposed to the public.

cameron-eyds: But yes i agree, Staff Filings that have no consumerReferenceId will not appear either way, so we can do without that condition i believe.

src/components/bcros/filing/CommonTemplate.vue
cameron-eyds: I think we will want the FF yes but i'm going to refactor to request on panel expand. We will only need it in that one place. 
The staff filing condition is probably redundant, was more of cautious coding but i think the way it's done now, we won't rending in absence of filingId === consumerReferenceId match, so it's probably safe to remove.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: > We will want to check if the current account is Staff. The DRS UI isn't directly exposed to the public.

I agree, but that's not what "isStaffFiling" does.

> But yes i agree, Staff Filings that have no consumerReferenceId will not appear either way, so we can do without that condition i believe.

Cool. And if you didn't load documents for non-staff, you wouldn't need to check that either ;)

severinbeauvais: You decide if you want to update anything. I'm OK with this as-is.

severinbeauvais (approved):
LGTM. Cam, merge when you're happy with it.

severinbeauvais:
Oops, you need a quick rebase, and please double-check the app version.

cameron-eyds:
> LGTM. Cam, merge when you're happy with it.

Thanks sir, i have another commit, the refactors we discussed offline. Basically fetching on demand vs up front with the dashboard, works good. 
Will commit after kid pick up!

src/components/bcros/filing/common/HeaderActions.vue
```diff
@@ -369,5 +379,11 @@ const handleButtonClick = async () => {
 
     ui.fetchingData = false
   }
+
+  // If document records are enabled and the documents list is empty,
+  // fetch the document records for the current business identifier.
+  if (enableDocumentRecords.value && !documents.value?.length) {
+    await getCorpDocuments({ consumerIdentifier: currentBusinessIdentifier.value })
+  }
```
severinbeauvais, line 387: I think this should be moved to the block above, or at least within the spinner's scope. This will keep the spinner visible while documents are being loaded.

src/pages/dashboard.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { CorpTypeCd, FilingTypes } from '@bcrs-shared-components/enums'
+import { storeToRefs } from 'pinia'
```
severinbeauvais, line 3: No longer needed?

cameron-eyds: Was never needed actually it's a Nuxt auto import, editor probably brought it in. Redundant, will remove.

src/components/bcros/filing/common/HeaderActions.vue
cameron-eyds: This would tie it to the filings documents (outputs), ie no outputs, no request to DRS. I don't know it well enough to say thats always the case. 
So we probably want to increase the scope of the spinner?

severinbeauvais: Looks good.

severinbeauvais:
Try re-running the Cypress tests. They often pass the second time.

cameron-eyds:
> Try re-running the Cypress tests. They often pass the second time.

Unfortunately i don't have permissions on this repo.

severinbeauvais:
> > Try re-running the Cypress tests. They often pass the second time.
> 
> Unfortunately i don't have permissions on this repo.

You do now 😉

cameron-eyds:
> > > Try re-running the Cypress tests. They often pass the second time.
> > 
> > 
> > Unfortunately i don't have permissions on this repo.
> 
> You do now 😉

Gracias!

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-133-ww3dxzjc.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-133-ww3dxzjc.web.app

PR #138, opened 2025-02-03
25642 - Delete not needed NoW codes for ToDoList

ArwenQin:
*Issue:*https://github.com/bcgov/entity/issues/25642

*Description of changes:*
- Clean up:
Since there will be no NoW drafts now, for Business Dashboard: 
We remove the codes handling NoW in To Do List, and in doResumeFiling() action

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.0.14",
+  "version": "1.0.15",
   "scripts": {
     "build": "nuxt generate",
     "build:local": "nuxt build",
--- a/src/interfaces/tasks-i.ts
+++ b/src/interfaces/tasks-i.ts
@@ -52,7 +52,6 @@ export interface TaskToDoI {
   documents?: Array<any>
   header: TaskApiHeaderI
   incorporationApplication?: any
-  noticeOfWithdrawal?: any
   registrarsNotation?: any
   registrarsOrder?: any
   registration?: any
--- a/src/interfaces/todo-i.ts
+++ b/src/interfaces/todo-i.ts
@@ -78,7 +78,4 @@ export interface TodoItemI {
   submitter?: string
   submittedDate?: Date
   latestReviewComment?: string
-
-  // Notice of Withdrawal only
-  filingToBeWithdrawn?: number
 }
--- a/src/utils/todo/action-functions.ts
+++ b/src/utils/todo/action-functions.ts
@@ -145,16 +145,6 @@ export const doResumeFiling = (item: TodoItemI): void => {
       params = { 'conversion-id': item.filingId.toString() }
       break
 
-    case FilingTypes.NOTICE_OF_WITHDRAWAL:
-      // navigate to Notice of Withdrawal page of Filings UI
-      navigateFn = goToFilingsUI
-      path = `/${currentBusinessIdentifier}/notice-of-withdrawal`
-      params = {
-        filingToBeWithdrawn: item.filingToBeWithdrawn.toString(),
-        filingId: item.filingId.toString()
-      }
-      break
-
     case FilingTypes.SPECIAL_RESOLUTION:
       // navigate to Edit UI to resume this Special Resolution
       navigateFn = goToEditUI
--- a/src/utils/todo/task-filing/content-loader.ts
+++ b/src/utils/todo/task-filing/content-loader.ts
@@ -114,8 +114,6 @@ export const getDraftTitle = (filing: TaskToDoI): string => {
       return filingTypeToName(FilingTypes.DISSOLUTION)
     case FilingTypes.INCORPORATION_APPLICATION:
       return FilingNames.INCORPORATION_APPLICATION
-    case FilingTypes.NOTICE_OF_WITHDRAWAL:
-      return FilingNames.NOTICE_OF_WITHDRAWAL
     case FilingTypes.REGISTRATION:
       return FilingNames.REGISTRATION
     case FilingTypes.RESTORATION:
--- a/src/utils/todo/task-filing/index.ts
+++ b/src/utils/todo/task-filing/index.ts
@@ -133,11 +133,6 @@ export const buildFilingTodo = async (task: TaskI): Promise<TodoItemI> => {
     newTodo.comment = filingData.comment
   }
 
-  // For NoW filing, add relevant fields to newTodo
-  if (isFilingType([FilingTypes.NOTICE_OF_WITHDRAWAL])) {
-    newTodo.filingToBeWithdrawn = filing.noticeOfWithdrawal.filingId
-  }
-
   // Add the actionButton for newTodo
   addActionButton(newTodo)
 


severinbeauvais (commented):
If there is a payment issue, it's possible that the NOW will appear as a task in the To Do list. (We have seen this with Court Orders -- you can search for a ticket on that.)

Anyway, I'm in favour of cleaning up unused task/todo properties _as long as there is default handling for the NOW if it somehow ends up in the To Do list_. You may have to test to see what happens.
- we should not display a Resume action
- we should display a Delete action

I will review further after you confirm what happens (with the changes in this PR) when a NOW happens to end up in the To Do list.

ArwenQin:
@severinbeauvais 
Thank you Severin.
1. For regular business, when payment is incomplete:
The draft goes to the ToDoList
image
Since the draft can't be resumed, the user must delete this draft, and then they can start another NoW.
**I can update to display a Delete action, instead of Resume action, and remove the drop down menu.**
(https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0887257?filing_id=169750&accountid=3040)


2. For temp business, it's a bit tricky:
When payment is incomplete, the NoW draft is saved and gets embedded, and it goes to the Filing History List.
The temp business dashboard currently doesn't display (doesn't handle) anything in ToDoList
image
(The label "Filed and Paid" is incorrect)
But it's actually payment incomplete; it must be deleted; otherwise the user can't file another NoW.
My thoughts:
a. Handle the NoW draft in ToDoList, which is new but I believe it's possible. However, NoW will be the only item in temp dashboard's ToDoList.
b. Keep it in the filing History list, and display a Delete action, instead of "Show Document", remove the drop down menu.
(https://dev.business-dashboard.bcregistry.gov.bc.ca/TYX0y2CKiN?accountid=3040)


Found a bug in Filing UI: 
First, enter some invalid BC online account information, the click Submit;
Show the Pay error message;
Switch to "No Pay", try to submit again,  trigger the error:
image
This is because I recently changed the codes, and only allow filingId = 0 (a new draft). The first Submit created a draft, and the second Submit will file the old draft (filing !=0). I will fix this issue in Filing UI.

cc: @meawong @eason-pan-bc

ArwenQin:
Implemented for regular filing:
image

severinbeauvais:
Can you update the code so that any default (unknown/unexpected) tasks in the To Do list show only a Delete action (no Resume)?

I am assuming that all _known_ tasks follow a different code path and display both a Resume and Delete button. Please let me know if this is not the case.

ArwenQin:
> Can you update the code so that any default (unknown/unexpected) tasks in the To Do list show only a Delete action (no Resume)?
> 
> I am assuming that all _known_ tasks follow a different code path and display both a Resume and Delete button. Please let me know if this is not the case.

@severinbeauvais 
This is the condition when to show "delete draft" only
image

eason-pan-bc:
Since for NoW, if the payment fails, users can only delete the draft. 

Would the idea below be a good option? 
- Delete that draft automatically,
- Show a message on the dashboard, noticing users that the payment failed, please file the NoW again.

ArwenQin:
> Since the embedded NOW needs to be extracted from the FE filing anyway, can you look at its status and conditionally add the NOW to the To Do list or the Filing History list?

I think it's possible, trying now

severinbeauvais:
> Since for NoW, if the payment fails, users can ONLY delete the draft. They cannot open / edit / resume the draft.
> 
> Would the idea below be a good option?
> 
> * Delete that draft automatically,
> * Show a message on the dashboard, noticing users that the payment failed, please file the NoW again.

I don't like the idea of the UI automatically doing something like this. If there was a precedent then it would be different. Thoughts?

severinbeauvais:
> @severinbeauvais This is the condition when to show "delete draft" only

I do not know this code inside-out... but I would like to see "known" tasks/filings handled explicitly, with a reasonable fallback for unknown items.

In this example, all known filings are not listed. Maybe they use the default result? I would prefer the default result to be "return true" and all other filings returning false accordingly.

I think this design will be easier to manage (in future) and provide safer handling for any new filings or filings that find themselves in an unexpected status (like the NOW, court order, etc, which don't support draft resumption).

Having said this, how much work is it to implement this update?

eason-pan-bc:
> I don't like the idea of the UI automatically doing something like this. If there was a precedent then it would be different. Thoughts?

My logic was about although there is a button letting users to click, but users don't actually have choices. (it's like you can click Yes or click Yes)
But, I agree with you on the consistency of UI behaviours. And keep it transparent to users and let users make the call.

ArwenQin:
> > @severinbeauvais This is the condition when to show "delete draft" only
> 
> I do not know this code inside-out... but I would like to see "known" tasks/filings handled explicitly, with a reasonable fallback for unknown items.
> 
> In this example, all known filings are not listed. Maybe they use the default result? I would prefer the default result to be "return true" and all other filings returning false accordingly.
> 
> I think this design will be easier to manage (in future) and provide safer handling for any new filings or filings that find themselves in an unexpected status (like the NOW, court order, etc, which don't support draft resumption).
> 
> Having said this, how much work is it to implement this update?

Got it! Not much work, I will do it now

severinbeauvais:
> Got it! Not much work, I will do it now

Great! Do you know if all known filings (that can be normally resumed) will need to be listed with "return false"?

ArwenQin:
> > Got it! Not much work, I will do it now
> 
> Great! Do you know if all known filings (that can be normally resumed) will need to be listed with "return false"?

Yes, I can refer to the doResumeFiling function, where listed all known filing types

src/utils/todo/task-filing/button-loader.ts
```diff
@@ -110,12 +110,26 @@ const showDeleteOnly = (todoItem: TodoItemI): boolean => {
...
+    case FilingTypes.REGISTRATION:
+    case FilingTypes.CHANGE_OF_REGISTRATION:
+    case FilingTypes.CONVERSION:
+    case FilingTypes.RESTORATION:
+      return false
     case FilingTypes.ALTERATION:
     case FilingTypes.DISSOLUTION:
     case FilingTypes.SPECIAL_RESOLUTION:
       return business && !business.currentBusiness.goodStanding && !isStaffAccount
     default:
```
ArwenQin, line 131: Those are the 13 known filing types listed in the doResumeFiling function, which means they can resume a draft

severinbeauvais: Wonderful, thanks!

src/components/bcros/todo/Item.vue
```diff
@@ -88,7 +89,13 @@ const handleClick = (button: ActionButtonI) => {
       // open the dialog for confirming deleting a draft filing (for existing businesses)
       if (businessId) { confirmDialog.value = confirmDeleteDraft }
       // open the dialog for confirming deleting a draft application (for temp business number)
-      if (bootstrapIdentifier.value) { confirmDialog.value = confirmDeleteApplication }
+      if (bootstrapIdentifier.value) {
+        if (prop.item.name === FilingTypes.NOTICE_OF_WITHDRAWAL) {
```
severinbeauvais, line 93: Is there a mixin or utility function that you could use instead?

severinbeauvais: PS Good catch. I agree it should say "delete draft".

ArwenQin:
Successfully add the draft (Payment incomplete) NoW to the ToDoList in temp business dashoboard:
image

**However, delete this filing will delete the whole temp business (will delete the original FE IA filing)**, I think we need an update in the end point to handle delete this NoW filing.
In addition, probably update back end to re-set the withdrawalPending property, remove the embedded NoW.
@meawong @eason-pan-bc

src/components/bcros/todo/Item.vue
severinbeauvais: Actually, should this be like the other code where you specify which filings show "delete application" and the rest use a default (delete draft)?

I was wondering about this when thinking about payment-error court-order filings...

src/stores/filings.ts
```diff
@@ -158,29 +158,38 @@ export const useBcrosFilings = defineStore('bcros/filings', () => {
...
+            order: noticeOfWithdrawal.noticeOfWithdrawal.courtOrder
+          },
+          latestReviewComment: header.latestReviewComment
+        } as ApiResponseFilingI)
+      }
+      else{useBcrosTodos().loadBootstrapTask({
+        enabled: true,
+        order: 0,
+        task: { filing: noticeOfWithdrawal }
+      } as TaskI)}
```
JazzarKarim, line 192: Please if you can fix the lint errors here!

severinbeauvais:
> **However, delete this filing will delete the whole temp business (will delete the original FE IA filing)**, I think we need an update in the end point to handle delete this NoW filing. In addition, probably update back end to re-set the withdrawalPending property, remove the embedded NoW. @meawong @eason-pan-bc

Why is the To Do item not titled as a NOW filing?

Doesn't the To Do item have its own filing id for deletion?

src/components/bcros/todo/Item.vue
ArwenQin: Previously, delete a draft in a temp business will delete the whole business, and redirect to My Business Registry page. I believe that's why for Bootstrap, it uses "delete application".
Now we probably need to implement a new endpoint, which deletes the NoW draft only

ArwenQin:
> > **However, delete this filing will delete the whole temp business (will delete the original FE IA filing)**, I think we need an update in the end point to handle delete this NoW filing. In addition, probably update back end to re-set the withdrawalPending property, remove the embedded NoW. @meawong @eason-pan-bc
> 
> Why is the To Do item not titled as a NOW filing?
> 
> Doesn't the To Do item have its own filing id for deletion?

I will fix the title label.
The To Do item have its own filing id, and I checked the url, it's correct. Like
deleteDraft url: https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v2/businesses/Tuy6V2Hfrh/filings/169864
for the NoW filing in https://dev.business-dashboard.bcregistry.gov.bc.ca/Tuy6V2Hfrh?accountid=3040
The NoW's filing id is 169864, which is correct.
But the it will delete the whole business

src/components/bcros/todo/Item.vue
severinbeauvais: I think the current delete endpoint should work for the NOW filing.

The decision to redirect to the MBR/BRD page is done by the dashboard. In your situation, no redirect is needed, although you will want to reload/refresh all the data so that the NOW disappears from the To Do list.

severinbeauvais:
> The To Do item have its own filing id, and I checked the url, it's correct. Like deleteDraft url: https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v2/businesses/Tuy6V2Hfrh/filings/169864 for the NoW filing in https://dev.business-dashboard.bcregistry.gov.bc.ca/Tuy6V2Hfrh?accountid=3040 The NoW's filing id is 169864, which is correct. But the it will delete the whole business

Oh, yes, I see. The NOW filing is "attached" to the temp business id, and the delete endpoint isn't handling this special case. Could it simply be updated?

src/stores/todo.ts
```diff
@@ -96,7 +96,11 @@ export const useBcrosTodos = defineStore('bcros/todos', () => {
 
     if (newTodo) {
       // bootstrap cases need this overwritten
-      newTodo.title = useBcrosBusinessBootstrap().bootstrapFilingDisplayName
+      if (newTodo.draftTitle) {
+        newTodo.title = newTodo.draftTitle
+      } else {
```
ArwenQin, line 101: if there is a draft title, use it as the label
image

src/stores/filings.ts
ArwenQin: Thank you, resolved

ArwenQin:
> > The To Do item have its own filing id, and I checked the url, it's correct. Like deleteDraft url: https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v2/businesses/Tuy6V2Hfrh/filings/169864 for the NoW filing in https://dev.business-dashboard.bcregistry.gov.bc.ca/Tuy6V2Hfrh?accountid=3040 The NoW's filing id is 169864, which is correct. But the it will delete the whole business
> 
> Oh, yes, I see. The NOW filing is "attached" to the temp business id, and the delete endpoint isn't handling this special case. Could it simply be updated?

@eason-pan-bc @meawong 
cc Eason and Megan here for their knowledge in the end point. Thank you!

src/stores/filings.ts
```diff
@@ -184,12 +184,13 @@ export const useBcrosFilings = defineStore('bcros/filings', () => {
           },
           latestReviewComment: header.latestReviewComment
         } as ApiResponseFilingI)
+      } else {
+        useBcrosTodos().loadBootstrapTask({
+          enabled: true,
+          order: 0,
+          task: { filing: noticeOfWithdrawal }
+        } as TaskI)
```
severinbeauvais, line 192: Does the NOW only display in the To Do list when this is a bootstrap business/application?

Or would the NOW display in the Recent Filings list when it's paid or completed? (only display in To Do list if draft or pending)

ArwenQin: For regular business:
 - only display in To Do list when the NoW is in Draft status (When payment incomplete, it's in Draft)
 
image

 - display in Recent Filings list when it's Paid or Completed.
 Does "pending" mean Paid? When it's in Paid status, the NoW is pending
image


For bootstrap:
 - only display in To Do list when the NoW is in Draft status (When payment incomplete, it's in Draft)
 
image

 - display in Recent Filings list when it's Paid or Completed.
image

severinbeauvais: Looks great !!!

PENDING = waiting for payment to complete
PAID = waiting for Filer to complete

If you really want to know if it's an unsubmitted draft or if there was a payment attempt/failure, you have to look at payment status. I think it doesn't matter here.

meawong:
> Successfully add the draft (Payment incomplete) NoW to the ToDoList in temp business dashoboard: image
> 
> **However, delete this filing will delete the whole temp business (will delete the original FE IA filing)**, I think we need an update in the end point to handle delete this NoW filing. In addition, probably update back end to re-set the withdrawalPending property, remove the embedded NoW. @meawong @eason-pan-bc

Yes, after a quick look, I think we can update this logic in the delete filings endpoint: 

image

As for the latter points regarding the withdrawal_pending property and the embedded NoW, there is already logic in place to handle this if the NoW filing is deleted!

eason-pan-bc:
> > Successfully add the draft (Payment incomplete) NoW to the ToDoList in temp business dashoboard: image
> > **However, delete this filing will delete the whole temp business (will delete the original FE IA filing)**, I think we need an update in the end point to handle delete this NoW filing. In addition, probably update back end to re-set the withdrawalPending property, remove the embedded NoW. @meawong @eason-pan-bc
> 
> Yes, after a quick look, I think we can update this logic in the delete filings endpoint:
> 
> image
> 
> As for the latter points regarding the withdrawal_pending property and the embedded NoW, there is already logic in place to handle this if the NoW filing is deleted!

Sorry, just saw this, and yes, we would need to add a check for `filing_type != 'noticeOfWithdrawal`

ArwenQin:
Hide button for a PENDING NoW
image

ArwenQin:
Hide button for regular user:
image

src/utils/todo/task-filing/button-loader.ts
```diff
@@ -8,11 +8,16 @@ export const addActionButton = (todoItem: TodoItemI): void => {
...
+  // non-staff see no buttons for staff filings (cont out, conversion, correction, restoration, withdrawal)
   if (!isStaffAccount && isStaffTodo(todoItem)) {
     return
   }
 
+  // don't show buttons for PENDING NoW filings
+  if (isFilingStatusPendingNoW(todoItem)) {
+    return
+  }
+
```
ArwenQin, line 20: Hide the buttons for PENDING NoW

src/utils/todo/task-filing/helper.ts
```diff
@@ -85,7 +85,8 @@ export const isStaffTodo = (item: TodoItemI | TaskApiHeaderI): boolean => {
   return item.name === FilingTypes.CONTINUATION_OUT ||
     item.name === FilingTypes.CONVERSION ||
     item.name === FilingTypes.CORRECTION ||
-    item.name === FilingTypes.RESTORATION
+    item.name === FilingTypes.RESTORATION ||
+    item.name === FilingTypes.NOTICE_OF_WITHDRAWAL
 }
```
ArwenQin, line 90: Hide Button for regular user

severinbeauvais:
> Implemented for regular filing: image

Can you please check, in this example, whether there is a payment record? If there is then you probably cannot simply delete this draft.

ArwenQin:
> Can you please check, in this example, whether there is a payment record? If there is then you probably cannot simply delete this draft.
image

It has a paymentStatusCode "BCOL_ERROR", may I know your suggestions?

severinbeauvais:
> image
> 
> It has a paymentStatusCode "BCOL_ERROR", may I know your suggestions?

Yes, your scenario is different than if a payment record was created. Example (where I cancelled out of CC payment):

image

If Payment Status Code is Created (and/or there is a Payment Token), then this task cannot simply be deleted. For NOW, we decided that this would be an Ops ticket to resolve. For this PR, just make sure there is no Delete option.

ArwenQin:
> If Payment Status Code is CREATED (and/or there is a Payment Token), then this task cannot simply be deleted. For NOW, we decided that this would be an Ops ticket to resolve. For this PR, just make sure there is no Delete option.


Got it! I think when there is a payment token, the filing should be in status PENDING, and this situation is handled, made sure there is no DELETE action button.
I also added a check for the payment token, just to be safe

severinbeauvais:
I haven't reviewed everything but I'm satisfied with the requirements we talked about here and the screenshots. I'm leaving it to rest of the reviewers to verify that the code is satisfactory.

JazzarKarim (approved):
So I reviewed the code and all the comments, I think it looks good. Great job!!

ArwenQin:
@patrickpeinanw @ketaki-deodhar @Rajandeep98 @hfekete Do you want to take a review before merging it? Thanks very much.

src/components/bcros/todo/Item.vue
severinbeauvais: I have nothing more to add to this conversation. Go ahead and resolve it if there's nothing else to say here.

severinbeauvais (approved):

ketaki-deodhar (approved):

ArwenQin:
@severinbeauvais @JazzarKarim Could you please merge this PR, thanks a lot!

PR #142, opened 2025-02-11
25697 - updates to display Court Order receipt in ledger

Rajandeep98:
*Issue:*https://github.com/bcgov/entity#25697

*Description of changes:*

display Court Order receipt in ledger


Rajandeep98:
image

Rajandeep98:
backend is not yet merged therefore with dev it may not show reciept etc

Rajandeep98:
image

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-142-y71q6o0a.web.app

pnpm-lock.yaml
severinbeauvais, line 1: This file has way too many changes. I don't know why you get these but others done.

Anyways, please don't commit this lock file as-is.

src/components/bcros/filing/common/HeaderActions.vue
```diff
@@ -63,7 +63,7 @@
         <strong v-if="!isExpanded">{{ $t('button.filing.actions.requestACopy') }}</strong>
         <strong v-else>{{ $t('button.filing.actions.close') }}</strong>
       </template>
-      <template v-else-if="isTypeStaff">
+      <template v-else-if="isTypeStaff && !(FilingTypes.COURT_ORDER)">
```
severinbeauvais, line 66: Would it make more sense to make Court Orders not a type of staff filing?

src/utils/filings.ts
```diff
@@ -10,13 +10,16 @@ export const isFilingType =
...
-    isFilingType(filing, FilingTypes.COURT_ORDER) ||
     isFilingType(filing, undefined, FilingSubTypeE.DISSOLUTION_ADMINISTRATIVE) ||
     isFilingType(filing, FilingTypes.PUT_BACK_ON) ||
     isFilingType(filing, FilingTypes.REGISTRARS_NOTATION) ||
     isFilingType(filing, FilingTypes.REGISTRARS_ORDER)
 }
 
+export const isCourtOrderOrNotStaffFiling = (filing: ApiResponseFilingI) => {
+  return isFilingType(filing, FilingTypes.COURT_ORDER)
+}
```
severinbeauvais, line 21: Where is this used?
How does this return True when it's not a staff filing?
Did you check everywhere that isStaffFiling is used and confirm that it doesn't affect court order? (In other words, whether a court order filing _should_ return True in some places?)

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-142-y71q6o0a.web.app

src/utils/download-file.ts
```diff
@@ -73,7 +73,7 @@ export const loadDocumentList = async (filing: ApiResponseFilingI) => {
           }
         } else if (groupName === 'uploadedCourtOrder') {
           const fileNumber = filing.data?.order?.fileNumber || unknownStr
-          const title = isStaffFiling(filing) ? `${filing.displayName} ${fileNumber}` : `${filing.displayName}`
+          const title = fileNumber ? `${filing.displayName} ${fileNumber}` : `${filing.displayName}`
```
severinbeauvais, line 76: `fileNumber` will always be truthy due to the fallback unknownStr.

src/components/bcros/filing/CommonTemplate.vue
```diff
@@ -70,7 +70,7 @@
       <slot name="documents">
         <!-- if we have documents, show them -->
         <!-- NB: staff filings don't have documents - see StaffFiling.vue for any exceptions -->
-        <template v-if="!isStaffFiling(filing) && filing.documentsLink">
+        <template v-if="!isStaffFiling(filing) && !isCourtOrderType(filing) && filing.documentsLink">
```
severinbeauvais, line 73: Does this mean that court order filing can't use this template?

Maybe court order should have its own template, like the other non-staff filings. Thoughts?

Rajandeep98:
image

severinbeauvais:
Reviewers, please review. Thanks!

src/components/bcros/filing/CommonTemplate.vue
Rajandeep98: its already using the existing one, but this condition is calling the default template as well, so showing the same thing twice. the existing staff template already has what we need  this is just to use the same workflow how staff filing is using for showing documents

JazzarKarim (approved):
Please update the package.json version number. Also, please don't forget to squash and merge the PR.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: Oh, I get it. It's because the staff filing template and this template have duplicate components.

This is a good reason for court order to have its own template (even if it looks a lot like the staff template minus the BcrosFilingCommonDocumentsList below).

If left as-is then I think it might be confusing that court order uses staff template -- or at least it would be one more special case for developers to know -- which, IMO, is undesirable.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-142-y71q6o0a.web.app

pnpm-lock.yaml
severinbeauvais: I don't know if there's anything we can do to retain the original sub-dependency versions. For now, I am OK with their auto-update ... hopefully we find any bugs through testing.

severinbeauvais: Resolving.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-142-y71q6o0a.web.app

Rajandeep98:
image
with new tempplate

pnpm-lock.yaml
severinbeauvais, line 1: If the next person to PR to this repo has a lock file that undoes most of the updates below, then I'm inclined to think you are the one with the different process or tool versions. Theoretically, any of us doing a `pnpm i` should end up with the same lock file. If  yours is different then let's find out why so we don't keep committing vastly different lock files.

src/utils/download-file.ts
severinbeauvais: Fixed. Resolving.

src/components/bcros/filing/CommonTemplate.vue
severinbeauvais: OK, this looks awesome to me. Much less confusion. Thanks!!!

severinbeauvais (approved):
Please fix broken unit tests and then I approve merging this PR.

pnpm-lock.yaml
severinbeauvais: I changed my mind.

I checked out this branch locally and ran `pnpm i` and was informed that THIS file is corrupt (for example, see duplicate lines 31 and 32, and there are many more like this). Fortunately, `pnpm i` created a non-correct lock file.

Do not merge with THIS file as-is. This needs to be a valid file.

severinbeauvais: E.g.,

image

severinbeauvais: Resolution: I pushed up a correct pnpm lock file to this branch. In fact, there are no changes needed to the lock file (it's the same as main branch). The unit tests now pass. Resolving this conversation.

severinbeauvais (approved):
LGTM. Just need to update the app version and then we can squash and merge this PR.

PR #150, opened 2025-02-24
26070  redirect to brd

Rajandeep98:
*Issue:*bcgov/entity#26070

*Description of changes:*
redirection to brd

### DIFF
--- a/src/components/bcros/todo/Item.vue
+++ b/src/components/bcros/todo/Item.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { FilingTypes } from '@bcrs-shared-components/enums'
 import { filingTypeToName } from '~/utils/todo/task-filing/helper'
+import { getStaffDashCrumb } from '~/utils/breadcrumbs'
 const t = useNuxtApp().$i18n.t
 const todosStore = useBcrosTodos()
 const { currentBusinessIdentifier, currentBusinessName } = storeToRefs(useBcrosBusiness())
@@ -12,7 +13,7 @@ const hasCancelPaymentError = ref(false)
 const confirmDialog = ref<DialogOptionsI | null>(null)
 const { redirect } = useBcrosNavigate()
 const emit = defineEmits(['expand', 'reload'])
-
+const breadcrumb = getStaffDashCrumb()
 const prop = defineProps({
   item: { type: Object as PropType<TodoItemI>, required: true },
   expanded: { type: Boolean, required: true }
@@ -145,7 +146,7 @@ const deleteApplication = async (): Promise<void> => {
       redirect(runtimeConfig.public.authWebURL)
     } else {
       // go to BCROS home page
-      redirect(runtimeConfig.public.businessesURL)
+      redirect(breadcrumb.href)
     }
   })
 }
--- a/src/components/bcros/todo/Item.vue
+++ b/src/components/bcros/todo/Item.vue
@@ -140,15 +140,8 @@ const deleteApplication = async (): Promise<void> => {
     // this logic does not exist in the old codebase.
     if (hasDeleteError.value) { return }
 
-    // N.B.: in '.env.example', authWebURL and businessesURL are the same
-    if (prop.item.nameRequest) {
-      // go to My Business Registry page
-      redirect(runtimeConfig.public.authWebURL)
-    } else {
-      // go to BCROS home page
-      console.log(breadcrumb.href)
-      redirect(breadcrumb.href)
-    }
+    // go to BRD
+    redirect(breadcrumb.href)
   })
 }
 
--- a/src/components/bcros/todo/Item.vue
+++ b/src/components/bcros/todo/Item.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { FilingTypes } from '@bcrs-shared-components/enums'
 import { filingTypeToName } from '~/utils/todo/task-filing/helper'
-import { getStaffDashCrumb } from '~/utils/breadcrumbs'
+import { getRegistryDashCrumb } from '~/utils/breadcrumbs'
 const t = useNuxtApp().$i18n.t
 const todosStore = useBcrosTodos()
 const { currentBusinessIdentifier, currentBusinessName } = storeToRefs(useBcrosBusiness())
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-150-vd4s3lrj.web.app

Rajandeep98:
image
redirection : https://dev.business-registry-dashboard.bcregistry.gov.bc.ca/account/3040?accountid=3040

src/components/bcros/todo/Item.vue
```diff
@@ -145,7 +146,8 @@ const deleteApplication = async (): Promise<void> => {
       redirect(runtimeConfig.public.authWebURL)
     } else {
       // go to BCROS home page
-      redirect(runtimeConfig.public.businessesURL)
+      console.log(breadcrumb.href)
```
severinbeauvais, line 149: Remove this statement before merging.

Also, I don't understand why the conditional on line 144 exists... or why there are separate redirects...? Try to find out if this is needed. If not then you can simplify this even more :)

severinbeauvais: Did you ask Hrvoje or Patrick why this was original implemented like this?

I'm OK with the code removal / simplification as long as we're not missing a use case here (and I'm not sure).

Rajandeep98: @hfekete @patrickpeinanw 
hello , Please verify this edge case  information for us ty

JazzarKarim (approved):
I tested in the temp link, it works. Code LGTM.

src/components/bcros/todo/Item.vue
```diff
@@ -12,7 +13,7 @@ const hasCancelPaymentError = ref(false)
 const confirmDialog = ref<DialogOptionsI | null>(null)
 const { redirect } = useBcrosNavigate()
 const emit = defineEmits(['expand', 'reload'])
-
+const breadcrumb = getRegistryDashCrumb()
```
hfekete, line 16: Just a note, not required change.

I would move this additionally to the line 142, as you are not using it anywhere else, other then for redirect

src/components/bcros/todo/Item.vue
```diff
@@ -139,14 +140,8 @@ const deleteApplication = async (): Promise<void> => {
...
-    // N.B.: in '.env.example', authWebURL and businessesURL are the same
-    if (prop.item.nameRequest) {
-      // go to My Business Registry page
-      redirect(runtimeConfig.public.authWebURL)
-    } else {
-      // go to BCROS home page
-      redirect(runtimeConfig.public.businessesURL)
-    }
+    // go to BRD
+    redirect(breadcrumb.href)
```
hfekete, line 144: If you are sure that you will always redirect user on deletion of the application to the dashboard href I believe that this simplification is fine. 
Though, I think both link will end up redirecting to the same page with a bit different looking links:

in your version it is going e.g.:
https://dev.business-registry-dashboard.bcregistry.gov.bc.ca/account/3101?accountid=3101

in the original it would go to:
https://dev.account.bcregistry.gov.bc.ca/decide-business?accountid=3101

And I am not sure both pages are the same. 

If the new redirects are what should happen and is tested for both `todo name request` and at least one other `todo item` I would guess new changes are OK.

hfekete (approved):
Approving with a comment.
Please see if you agree with comments. 

if you agree, code LGM.

Cheers!

src/components/bcros/todo/Item.vue
kialj876: you guys put this code in yourselves I think haha
https://github.com/bcgov/business-dashboard-ui/pull/148/files

severinbeauvais: Oh, no. We restored some code that was previously deleted.

Here is Patrick's PR that added this code: https://github.com/bcgov/business-dashboard-ui/pull/44

src/components/bcros/todo/Item.vue
severinbeauvais: What is a "todo name request"?

severinbeauvais: Is the expectation that a NR would be a todo item, and could be deleted?

It's looking like this code was invalid, since a NR is not an item in the To Do list and therefore cannot be deleted, unless this was changed. (A draft IA may be in the To Do list and contain a NR object.)

Rajan will see if she can test this NR flow. If it's not used then we will proceed with simplifying this redirect.

src/components/bcros/todo/Item.vue
severinbeauvais: Resolving this conversation as there's a more recent one on the same subject below.

Rajandeep98:
image

Rajandeep98:
currently redirecting to brd for nr ia as well.

src/components/bcros/todo/Item.vue
severinbeauvais: @Rajandeep98 , please respond to this comment.

Rajandeep98: keeping it for future usage

Rajandeep98:
> > currently redirecting to brd for nr ia as well.
> 
> OK, good.
> 
> Does the old code (checking prop.item.nameRequest) make any sense?

https://dev.account.bcregistry.gov.bc.ca/
hitting this in both cases from config view, so doesnt matter

severinbeauvais:
> > Does the old code (checking prop.item.nameRequest) make any sense?
> 
> https://dev.account.bcregistry.gov.bc.ca/ hitting this in both cases from config view, so doesnt matter

That's not my question. I was asking if it made any sense to check the nameRequest proper for any reason.

Also, what is the config view?

severinbeauvais:
**Summary of my concern around different handling if the todo item has a NR:**

The code in this repo mirrored the original code in Filings UI (old dashboard). Unfortunately, the Filings UI code was also missing code comments explaining why the redirect URL was different if a draft application has a NR.

But I can guess: if there was a NR then the intent might have been to redirect the user to their MBR/BRD so they could see the NR and do something else with it. Or, if it was a numbered draft application then the intent might have been to redirect the user to the "Start a B.C. based Business..." page where more instructions are provided.

I'm OK with the simplification that the user is redirected to the MBR when they delete either a named or numbered draft application.

severinbeauvais (approved):

PR #160, opened 2025-03-05
25649 remove deprecated ff

TVWerdal:
*Issue:*https://github.com/bcgov/entity/issues/25649

Removed explicit calls to the "supported-business-summary-entities". Left in the cases for AllowableActionE.BUSINESS_SUMMARY - I could not see this used anywhere else, but I was worried stripping it may cause issues. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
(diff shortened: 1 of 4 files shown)


TVWerdal:
Got a more updated version of this, closing this PR to replace it with that one.

PR #164, opened 2025-03-06
26368 - After a COA is withdrawn, don't show Pending badge

ArwenQin:
*Issue:*https://github.com/bcgov/entity/issues/26368

*Description of changes:*
- After a COA is withdrawn, it shouldn't show the Pending badge and tooltip
- Fix the pendingAddress condition to exclude withdrawn COA

### DIFF
--- a/src/pages/dashboard.vue
+++ b/src/pages/dashboard.vue
@@ -252,20 +252,13 @@ const alerts = computed((): Array<Partial<AlertI>> => {
 })
 
 const pendingAddress = computed(() => {
-  const currentDate = new Date()
-  if (filings && filings.value && filings.value.length > 0) {
-    const coaFilings = filings.value.filter((filing) => {
-      return (filing.name === FilingTypes.CHANGE_OF_ADDRESS) && (filing.status !== FilingStatusE.WITHDRAWN)
-    })
-    coaFilings.sort((a, b) => {
-      return new Date(b.effectiveDate) - new Date(a.effectiveDate)
-    })
-    const coaFiling = coaFilings[0]
-    if (coaFiling?.effectiveDate && new Date(coaFiling?.effectiveDate) > currentDate) {
-      return true
-    }
-  }
-  return false
+  if (!filings?.value?.length) { return false }
+
+  // Return true if the COA is pending
+  return filings.value.some(filing =>
+    filing.name === FilingTypes.CHANGE_OF_ADDRESS &&
+    filing.status === FilingStatusE.PAID
+  )
 })
 
 const showChangeOfAddress = ref(false)
--- a/src/pages/dashboard.vue
+++ b/src/pages/dashboard.vue
@@ -255,7 +255,7 @@ const pendingAddress = computed(() => {
   const currentDate = new Date()
   if (filings && filings.value && filings.value.length > 0) {
     const coaFilings = filings.value.filter((filing) => {
-      return filing.name === FilingTypes.CHANGE_OF_ADDRESS
+      return (filing.name === FilingTypes.CHANGE_OF_ADDRESS) && (filing.status !== FilingStatusE.WITHDRAWN)
     })
     coaFilings.sort((a, b) => {
       return new Date(b.effectiveDate) - new Date(a.effectiveDate)
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.0.28",
+  "version": "1.0.29",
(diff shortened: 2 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-164-oo8p6dol.web.app

src/pages/dashboard.vue
```diff
@@ -255,7 +255,7 @@ const pendingAddress = computed(() => {
   const currentDate = new Date()
   if (filings && filings.value && filings.value.length > 0) {
     const coaFilings = filings.value.filter((filing) => {
-      return filing.name === FilingTypes.CHANGE_OF_ADDRESS
+      return (filing.name === FilingTypes.CHANGE_OF_ADDRESS) && (filing.status !== FilingStatusE.WITHDRAWN)
```
severinbeauvais, line 258: Which do you think is a better check:
- exclude withdrawn?
- include only pending?

JazzarKarim (approved):

ArwenQin:
Before fix:
image


After fix:
image


Test: 
https://business-dashboard-dev--pr-164-oo8p6dol.web.app/BC0887698

Pending COA (not withdrawn) is not impacted
https://business-dashboard-dev--pr-164-oo8p6dol.web.app/BC0887684

src/pages/dashboard.vue
ArwenQin: Can we include only pending?
Once the COA comes in effect, its status is Completed.
Only when it's pending, its status is Paid.
So we only include status === PAID.
And no need to check the date
image
Do you think it's safe?

severinbeauvais: Great questions.

Thank you for the correction. Yes, Paid instead of Pending.

On the one hand, if the COA is still Paid then it's technically still future effective.

On the other hand, if the COA was supposed to take effect at a particular time, and the Filer hasn't done its job yet, should we allow users to withdraw it? (Filer will back-date the COA to its intended effective time.) I don't know if there are legal ramifications.

Ask Mihai for his opinion, please.

ArwenQin: Currently, we don't allow users to withdraw a filing when its effective time has passed but the filer hasn't done processing it, because the File a Withdrawal button is disabled by checking the current Date > filing.effectiveDate.
It applies to all filing types those can be withdrawn.
@Mihai-QuickSilverDev

severinbeauvais: OK, then you could safely keep the date logic here instead of opening up the loophole (unless Mihai likes the loophole).

ArwenQin: but this const pendingAddress here only determines whether to show the Pending badge. I think it makes sense to show the Pending badge when the filer hasn't done processing it, even the date has passed.

ArwenQin: Just confirmed with Mihai, we don't allow filing a withdraw when the FE time passed
image

severinbeauvais: > but this const pendingAddress here only determines whether to show the Pending badge. I think it makes sense to show the Pending badge when the filer hasn't done processing it, even the date has passed.

Yes, that makes sense to me.

src/pages/dashboard.vue
```diff
@@ -252,20 +252,13 @@ const alerts = computed((): Array<Partial<AlertI>> => {
...
-  }
-  return false
+  if (!filings?.value?.length) { return false }
+
+  // Return true if the COA is pending
+  return filings.value.some(filing =>
+    filing.name === FilingTypes.CHANGE_OF_ADDRESS &&
+    filing.status === FilingStatusE.PAID
+  )
 })
```
ArwenQin, line 262: removed the check for date and the sort
@severinbeauvais

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-164-oo8p6dol.web.app

src/pages/dashboard.vue
severinbeauvais: To summarize, the beahviour without the date check is:
- if COA is pending and still future effective, then the badge shows and NOW is possible/enabled
- if COA is pending but in the past, then the badge shows and NOW is not possible/disabled, and it is not possible to start another filing

This seems OK to me. Please verify with Mihai.

ArwenQin: @Mihai-QuickSilverDev

Mihai-QuickSilverDev: - [ ] I am not sure what pending in the past means.
- [ ] Would need to meet with Arwen for a few minutes to discuss this live.

PR #165, opened 2025-03-07
24822 adding businessid value to query string

TVWerdal:
*Issue:*https://github.com/bcgov/entity/issues/24822

Adding businessid value to url.

### DIFF
--- a/src/composables/useBcrosNavigate.ts
+++ b/src/composables/useBcrosNavigate.ts
@@ -8,8 +8,12 @@ export const useBcrosNavigate = () => {
     // get account id and set in params
     const redirectURL = new URL(url)
     const accountId = account.currentAccount.id
+    const businessId = business.currentBusiness?.identifier
     if (accountId) {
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-165-xlsgec51.web.app

src/composables/useBcrosNavigate.ts
```diff
@@ -8,9 +8,13 @@ export const useBcrosNavigate = () => {
     // get account id and set in params
     const redirectURL = new URL(url)
     const accountId = account.currentAccount.id
+    const businessId = business.currentBusiness?.identifier
```
severinbeauvais, line 11: Seeing as line 51 doesn't use conditional chaining for the identifier, you don't need it either.

I prefer to not use conditional chaining unless it's needed -- and then, you need to use carefully -- because it can mask errors that you should be handling. It's a double-edged sword.

severinbeauvais (approved):

PR #167, opened 2025-03-10
26289 - Business Dashboard: should handle error if business id is invalid

Rajandeep98:
*Issue:*bcgov/entity#26289

*Description of changes:*
Business Dashboard: should handle error if business id is invalid
redirect to staff dashboard

### DIFF
--- a/src/app/router.options.ts
+++ b/src/app/router.options.ts
@@ -12,7 +12,7 @@ export default <RouterConfig> {
         breadcrumbs: [getBcrosHomeCrumb, getRegistryDashCrumb, getBusinessDashCrumb],
         staffBreadcrumbs: [getStaffDashCrumb, getBusinessDashCrumb]
       },
-      beforeEnter(to, from, next) {
+      beforeEnter(to, _from, next) {
         // If identifier is missing
         if (!to.params.identifier) {
           next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
--- a/src/app/router.options.ts
+++ b/src/app/router.options.ts
@@ -11,15 +11,15 @@ export default <RouterConfig> {
         title: 'Business Dashboard',
         breadcrumbs: [getBcrosHomeCrumb, getRegistryDashCrumb, getBusinessDashCrumb],
         staffBreadcrumbs: [getStaffDashCrumb, getBusinessDashCrumb]
+      },
+      beforeEnter(to, _from, next) {
+        // If identifier is missing
+        if (!to.params.identifier) {
+          next({ name: RouteNameE.DASHBOARD, params: { identifier: ' ' } })
+        } else {
+          next()
+        }
       }
-      // beforeEnter(to, _from, next) {
-      //   // If identifier is missing
-      //   if (!to.params.identifier) {
-      //     next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
-      //   } else {
-      //     next()
-      //   }
-      // }
     },
     {
       name: RouteNameE.CRITICAL_ERRORS,
--- a/src/app/router.options.ts
+++ b/src/app/router.options.ts
@@ -1,18 +1,24 @@
 import type { RouterConfig } from '@nuxt/schema'
 
 export default <RouterConfig> {
-  // https://router.vuejs.org/api/interfaces/routeroptions.html#routes
-  // alternatively, could put this inside the setup for each page
   routes: _routes => [
     {
       name: RouteNameE.DASHBOARD,
-      path: '/:identifier',
+      path: '/:identifier?',
       component: () => import('~/pages/dashboard.vue').then(r => r.default || r),
       meta: {
         layout: 'business',
         title: 'Business Dashboard',
         breadcrumbs: [getBcrosHomeCrumb, getRegistryDashCrumb, getBusinessDashCrumb],
         staffBreadcrumbs: [getStaffDashCrumb, getBusinessDashCrumb]
+      },
+      beforeEnter(to, from, next) {
+        // If identifier is missing
+        if (!to.params.identifier) {
+          next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
+        } else {
+          next()
+        }
       }
     },
     {
--- a/src/app/router.options.ts
+++ b/src/app/router.options.ts
@@ -15,7 +15,7 @@ export default <RouterConfig> {
       beforeEnter(to, from, next) {
         // If identifier is missing
         if (!to.params.identifier) {
-          next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
+          next({ name: RouteNameE.CRITICAL_ERRORS, params: { identifier: 'default_identifier' } })
         } else {
           next()
         }
--- a/src/app/router.options.ts
+++ b/src/app/router.options.ts
@@ -15,7 +15,7 @@ export default <RouterConfig> {
       beforeEnter(to, _from, next) {
         // If identifier is missing
         if (!to.params.identifier) {
-          next({ name: RouteNameE.DASHBOARD, params: { identifier: ' ' } })
+          next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
         } else {
           next()
         }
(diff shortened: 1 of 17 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

src/stores/business.ts
```diff
@@ -95,6 +95,9 @@ export const useBcrosBusiness = defineStore('bcros/business', () => {
       .then(({ data, error }) => {
         if (error.value || !data.value) {
           console.warn('Error fetching business contacts for', identifier)
+          if (error.value?.statusCode === 404) {
+            useBcrosNavigate().goToBcrosDashboard()
+          }
```
severinbeauvais, line 100: 1. This only handles a specific "business contact" call error from Auth API. What about all the other possible errors when calling Legal API?

2. Isn't this error already handled by the code below (lines 101-105)? In fact, it looks like many calls in this module have error handling, but why isn't it actually displaying the error(s)?

Maybe I'm missing something. Please explain how this code handles an error if the business id is invalid.

severinbeauvais:
This code change results in the following, which still show an empty page (for just a second) and does not display any errors.

[screen-capture.webm](https://github.com/user-attachments/assets/f98b3c0f-bba2-40a6-b2c5-94cfe5768482)

src/stores/business.ts
Rajandeep98: accidentally added to wrong method, updating it to show error dialogs before redirection/no redirection

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

Rajandeep98:
image

redirecting to brd

severinbeauvais:
When my current account is not allowed to load the business. This is correct.

image

https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/BC0887749?accountid=3040

severinbeauvais:
When I try to load an invalid business ID.

image

https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/BCxxxxxxx?accountid=668

severinbeauvais:
When I don't specify the business ID in the URL.

-> I am redirected straight to the BRD. No error is displayed.

https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/?accountid=668

src/lang/en.json
```diff
@@ -224,6 +227,18 @@
...
           "text": "An error happened while fetching affiliation invitation."
         },
+        "loadBusinessFetchError": {
+          "title": {
+            "accessRestricted": "Access Restricted",
+            "invalidLink": "Invalid Link",
+            "pageNotFound": "Page not found"
+          },
+          "text": {
+            "accessRestricted": "You dont have permission to access this business. Go to your Business Registry Dashboard to request access to this business.",
```
severinbeauvais, line 237: "dont" should be "don't"

src/stores/business.ts
severinbeauvais: Outdated comment. Resolving.

severinbeauvais:
The code looks great!

I still want to test this. Please provide sample URLs and steps to reproduce the various scenarios. Thanks!

Rajandeep98:
reproduce:
1. nr invalid: 404
https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/BCxxxxxxx 
2. no access / unauth : 403/401
use bscs id to login may reproduce this error or other way is to keep 2 tabs open for same business, logout from one, remove token from storage error will show up 

https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/FM1064603

3. 5xx error: not able to access BE : may try it on local
https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/FM1064603

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

src/lang/en.json
severinbeauvais: Fixed. Resolving.

TVWerdal:
@Rajandeep98 - Tested the first two cases, but from what I understand to test the 3rd case (5XX response) I would need to have your code on my local, correct?

severinbeauvais:
> reproduce:
<snip>

What I'm seeing, while logged in as BCREG0020, in all cases, is the original "Unable to access business" dialog. Ideally, one of the more specific dialogs would be displayed. Any idea what is causing the original dialog to display instead of another one, for non-staff users?

However, while logged in as staff, I see the following.

> nr invalid: 404

image

> no access / unauth : 403/401

I can't reproduce this one. If my KC token is missing (eg, I am logged out) then I am redirected to a login page.

> 5xx error: not able to access BE : may try it on local

I was able to reproduce this one by setting my network offline during the business polling period. I think this is another edge case. I think, ideally, business polling errors should fail silently, and then polling should restart when the network is restored. This would mean monitoring the network status, which is way out of scope for this ticket. So, I think this is fine.

image

By the way, if I navigate away from an error page, then click back, I get an empty page. I guess the errors page is looking in the store for some errors values but they are no longer there. Please create a ticket to fix this in the future -- it would be nice to see either the same error as before (ie, static error page), or some default error page. You can probably reproduce this by going to the error URL directly: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/errors/entity/BC0887749

src/app/router.options.ts
```diff
@@ -1,18 +1,24 @@
...
       meta: {
         layout: 'business',
         title: 'Business Dashboard',
         breadcrumbs: [getBcrosHomeCrumb, getRegistryDashCrumb, getBusinessDashCrumb],
         staffBreadcrumbs: [getStaffDashCrumb, getBusinessDashCrumb]
+      },
+      beforeEnter(to, from, next) {
+        // If identifier is missing
+        if (!to.params.identifier) {
+          next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
```
severinbeauvais, line 18: Clever solution!

Instead of setting a "special value", could you reroute to a (new) error page directly?

severinbeauvais: After testing this, it looks fine. Resolving.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

severinbeauvais:
@Rajandeep98 , can you please fix the lint error (in this case, add a line before it to ignore the rule), the unit test errors, and the Cypress errors, before this can be merged?

@TVWerdal , do you have time to review and verify this PR? The other reviewers are ooo today.

severinbeauvais (approved):
Please fix CI issues, and wait for another review, before merging.

TVWerdal:
The code looks good to me, and the tests I ran looked good. Are the failed tests below a concern?

TVWerdal (approved):
Looks good, approved pending the testing fixes.

severinbeauvais:
> as previously app error dialogs were not implemented, app didn't needed auth token setup. as of now it is failing due to fake id fake token mock from cy, need to update the base cypress setup, moving this tt ahead and creating a new one to fix the cy setup to accommodate new changes in the code.

Rajan and I chatted about this. It looks like the Cypress tests are simply failing to mock/intercept some API calls, and it looks like actual API calls are being made instead. There should be an update for this soon.

severinbeauvais:
Great job fixing the Cypress tests! Can you please look at the remaining broken unit tests?

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-167-4s1gnd8i.web.app

JazzarKarim (approved):
Great job!

src/app/router.options.ts
```diff
@@ -1,18 +1,24 @@
...
       meta: {
         layout: 'business',
         title: 'Business Dashboard',
         breadcrumbs: [getBcrosHomeCrumb, getRegistryDashCrumb, getBusinessDashCrumb],
         staffBreadcrumbs: [getStaffDashCrumb, getBusinessDashCrumb]
+      },
+      beforeEnter(to, _from, next) {
+        // If identifier is missing
+        if (!to.params.identifier) {
+          next({ name: RouteNameE.DASHBOARD, params: { identifier: 'default_identifier' } })
```
severinbeauvais, line 18: This shows an URL of `https://business-dashboard-dev--pr-167-4s1gnd8i.web.app/default_identifier`.

How about setting the default identifier to "XXX" so it looks more like an error in the URL? Or other ideas?

severinbeauvais:
Thank you for all the fixes, Rajan! This really improves the error handling and testing.

As for changing the default identifier string, what do you think? If you agree, any one of us can update it in our next PR.

PR #171, opened 2025-03-14
26369 - Load State Filing on Dashboard Refresh 

meawong:
*Issue:*https://github.com/bcgov/entity/issues/26369 

*Description of changes:*
- The bug happens in an intermediate stage when the state filing is not updated, however, once you **reload** the page (and the state filing is updated), you see the correct status & reasoning shown (i.e. if you go to the corp in the ticket, you'll see the correct status: 
image
)
- Solution: force the state filing to be updated on **refresh** (when you click the button that appears on the toast) to ensure the status shown is up-to-date

https://github.com/user-attachments/assets/f09a9a1f-7535-4c7b-a05d-ec1fc9e1255d

### DIFF
--- a/src/pages/dashboard.vue
+++ b/src/pages/dashboard.vue
@@ -195,6 +195,7 @@ const reloadBusinessInfo = async () => {
 
   // reload business info using the force=true flag
   await loadBusinessInfo(true)
+  await business.loadStateFiling(true)
 }
 
 onBeforeMount(async () => {
--- a/src/pages/dashboard.vue
+++ b/src/pages/dashboard.vue
@@ -193,7 +193,7 @@ const reloadBusinessInfo = async () => {
   useBcrosFilings().clearFilings()
   // TO-DO: also need to clear the pending filing list (not yet implemented)
 
-  // reload business info using the force=true flag
+  // reload business info and state filing using the force=true flag
   await loadBusinessInfo(true)
   await business.loadStateFiling(true)
 }
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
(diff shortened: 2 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-171-wtkdblh7.web.app

JazzarKarim (approved):
LGTM and don't forget package version please.

src/pages/dashboard.vue
```diff
@@ -193,8 +193,9 @@ const reloadBusinessInfo = async () => {
   useBcrosFilings().clearFilings()
   // TO-DO: also need to clear the pending filing list (not yet implemented)
 
-  // reload business info using the force=true flag
+  // reload business info and state filing using the force=true flag
   await loadBusinessInfo(true)
+  await business.loadStateFiling(true)
```
severinbeauvais, line 198: What does "true" mean here, and what if there isn't a state filing?

meawong: using the force=true flag

severinbeauvais: OK, so it reloads it even if it was already loaded. Is that necessary?

And I see that getStateFiling() returns if there is no URL, so that's OK.

meawong: Right now without this change, if I were to file a Put Back On filing (to reverse the Admin Dissolution in the video), after I hit refresh, it still thinks the state filing is the Admin Dissolution. We need to force reload the state filing so that it updates with the new state filing (in this case the Put Back On).

severinbeauvais: Aha! Perfect explanation. I get it. Makes sense. Looks good. Etc :)
Resolving.

severinbeauvais (approved):

PR #185, opened 2025-04-03
27003 - Fix the linter

JazzarKarim:
*Issue:*https://github.com/bcgov/entity/issues/27003

*Description of changes:*
- Fixed eslint
- Fixed the issues that were captured by eslint once it was fixed

`pnpm run lint` should work now.


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-185-p07crqw4.web.app

meawong (approved):

cypress/e2e/components/todos/draft.cy.ts
```diff
@@ -208,7 +208,8 @@ context('TODOs -> Draft Filing', () => {
         .click()
         .wait('@getAnnualReportFiling')
         .its('request.url')
-        .should('include', `/${identifier}/annual-report?accountid=1&businessid=CP1002605&filingId=${filingId}&arFilingYear=${arYear}`)
+        .should('include', `/${identifier}/annual-report?accountid=1`)
+        .should('include', `&businessid=CP1002605&filingId=${filingId}&arFilingYear=${arYear}`)
```
severinbeauvais, line 212: The new code is not equivalent to the old code (ditto in similar change above).

The new code allows the URL parameters to be anywhere in the URL, which is not correct.

I suggest using another way to break up the string, or use `eslint-disable-next-line` judiciously.

cypress/e2e/error-flows/invalid-name-request.cy.ts
```diff
@@ -4,7 +4,7 @@ import { IncorporationApplicationNr } from '../../fixtures/name-requests/incorpo
 import { NameRequestStateE } from '../../../src/enums/name-request-states-e'
 
 context('Temporary business with invalid name request', () => {
-  let nameRequest = undefined
+  let nameRequest
```
severinbeauvais, line 7: This is equivalent to the old code.

image

Did the linter complain about undefined? How about using null instead? Or empty object?

package.json
```diff
@@ -54,6 +54,8 @@
     "@zadigetvoltaire/nuxt-gtm": "^0.0.13",
     "cypress": "^13.10.0",
     "cypress-axe": "^1.5.0",
+    "eslint": "^8.57.0",
+    "eslint-plugin-tailwindcss": "^3.17.4",
```
severinbeauvais, line 58: 👍

severinbeauvais: I didn't take the time to look, but is there a linter config file?

cypress/e2e/error-flows/invalid-name-request.cy.ts
JazzarKarim: Yup, linter was not happy about this. 

I could try initializing as null, I think that should be OK.

package.json
JazzarKarim: Yup, there is a config file: https://github.com/bcgov/business-dashboard-ui/blob/main/.eslintrc

cypress/e2e/components/todos/draft.cy.ts
JazzarKarim: This should be good now.

cypress/e2e/error-flows/invalid-name-request.cy.ts
JazzarKarim: Initialized as null. Linter was happy about it.

severinbeauvais (approved):

PR #189, opened 2025-04-09
25048 add staff link now always displayed for staff

TVWerdal:
*Issue:*https://github.com/bcgov/entity/issues/25048

Update so that we check account id query value against account id session value when setting up user on the dashboard. Assumed that session account id takes precedence. This addresses case where incorrect query param was overriding proper account id which caused issues with loading the proper account data.

### DIFF
--- a/src/stores/account.ts
+++ b/src/stores/account.ts
@@ -154,13 +154,17 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
   }
 
   /** Set the user account list and current account */
-  async function setAccountInfo (currentAccountId?: number) {
+  async function setAccountInfo (currentAccountId = NaN) {
     // Check if we have a currentAccountId and if it matches what is stored in our sessionStorage
-    if (!currentAccountId ||
+    if (isNaN(currentAccountId) ||
       (JSON.parse(sessionStorage.getItem(SessionStorageKeyE.CURRENT_ACCOUNT)) &&
       currentAccountId !== JSON.parse(sessionStorage.getItem(SessionStorageKeyE.CURRENT_ACCOUNT)).id)) {
       // try getting id from existing session storage
       currentAccountId = JSON.parse(sessionStorage.getItem(SessionStorageKeyE.CURRENT_ACCOUNT) || '{}').id
+      // refresh the page so that account based checks are rerun
+      const url = new URL(window.location)
+      url.searchParams.set('accountid', currentAccountId.toString())
+      window.location.assign(url.toString())
     }
     if (user.value?.keycloakGuid) {
       userAccounts.value = await getUserAccounts(user.value?.keycloakGuid) || []
--- a/src/stores/account.ts
+++ b/src/stores/account.ts
@@ -155,34 +155,41 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
     }
   }
 
-    /** Set the user account list and current account */
-    async function setAccountInfo (currentAccountId?: number) {
-      const queryAccountId = currentAccountId
-      if (!currentAccountId || isNaN(currentAccountId)) {
-        // try getting id from existing session storage
-        currentAccountId = JSON.parse(sessionStorage.getItem(SessionStorageKeyE.CURRENT_ACCOUNT) || '{}').id
-      }
-      if (user.value?.keycloakGuid) {
-        userAccounts.value = await getUserAccounts(user.value?.keycloakGuid) || []
-        if (userAccounts && userAccounts.value.length > 0) {
-          currentAccount.value = userAccounts.value[0]
-          if (currentAccountId && currentAccountId !== Number(currentAccount.value.id)) {
-            console.log(userAccounts.value.find(account=>account.id === currentAccountId))
-            // if previous current account id selection information available set this as current account
-            userAccounts.value.find(account => account.id === currentAccountId) ?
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-189-2jvfay4w.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-189-2jvfay4w.web.app

src/stores/account.ts
```diff
@@ -155,7 +155,10 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
 
   /** Set the user account list and current account */
   async function setAccountInfo (currentAccountId?: number) {
-    if (!currentAccountId) {
+    // Check if we have a currentAccountId and if it matches what is stored in our sessionStorage
+    if (!currentAccountId ||
```
severinbeauvais, line 159: This will not work as expected if the id is 0.

Also, what is the value of the id if it's not passed in -- undefined? Maybe it would be better to use this instead?
`async function setAccountInfo (currentAccountId = NaN)`

Then on this line you could check `isNaN(currentAccountId)`.

src/stores/account.ts
severinbeauvais, line 1: There's a fundamental question here that I'm struggling with: should the UI behave according to the account id in the URL, or according to the current logged-in account?

Case in point: when the Change Profile page loads, it should definitely use the account id provided as a URL parameter. (In fact, when account id is provided as a URL parameter, the underlying account is updated.)

Here, we don't want to override the fact that the current user is staff, but what should we do if there's a different account id in the URL? I think the best thing is to remove the mismatching URL parameter. What do you think?

src/stores/account.ts
TVWerdal: Good catch, and I like the fix. I'll make that update.

src/stores/account.ts
TVWerdal: Are you meaning when you choose to switch account here? 
image

Or a different way of changing account? The Switch account action immediately pushes the id to the session storage so this doesn't interfere with that. 

I have noticed that having an incorrect accountid will break heading over to the profile page - so maybe if we find there is a missmatch with session we replace the query id with the proper one?

severinbeauvais: I'm uncomfortable having the UI behave as staff when the URL account id asks for something else.

severinbeauvais: > Are you meaning when you choose to switch account here?

No. I think the original scenario was where a staff user (or really, any user), goes to the business dashboard with an account id in the URL that doesn't match the user's actual account id.

> I have noticed that having an incorrect accountid will break heading over to the profile page - so maybe if we find there is a missmatch with session we replace the query id with the proper one?

If the URL account id asks for a particular account id that doesn't match the logged-in account id then we have 2 choices:
- sync the actual account id to what the URL says, or
- sync the URL to the actual account id

Which one makes the most sense? In this case, I think the second. (On the Profile page, I think the first one makes the most sense.) I don't like different behaviour like this but ... What do you think?

TVWerdal: Yeah, I was writing something similar when you posted this. I was coming at it from a standpoint that the url is easily manipulated while the session storage was more difficult to manipulate - so if they don't match then the session should take precedence.

severinbeauvais: IMO, the URL is "the page you're requesting".

The UI response might be "you can't access this".

But I think this is a special case. The expectation might be that you "see the page as per the URL account id" (ie, masquerade as a different account) but that would be complicated as you'd be working with the current user's roles/authorizations and the new account's roles/authz. I think a better UI response is "I don't support that, here's the page according to your current account" and "let me fix that URL for you".

severinbeauvais: Which means, yes, use the current account id from the Keycloak token / session storage, but also update the URL to the actual current account id. I'd be OK with a redirect, here. In other words, "you can't see the page you asked for so here's the page I can show you".

severinbeauvais: Sorry about the anthropomorphism, hahaha.

src/stores/account.ts
```diff
@@ -154,20 +154,34 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
...
+
+        // If we set a new currentAccountId, change it in the URL and reload.
+        if (currentAccountId !== queryAccountId || currentAccountId !== Number(currentAccount.value.id)) {
+          // currentAccount takes precedent over any other number.
+          currentAccountId = Number(currentAccount.value.id)
+          const url = new URL(window.location.href)
+          url.searchParams.delete('accountid')
+          url.searchParams.set('accountid', currentAccountId.toString())
+          // Cypress is not happy if we reload the page, so avoid that here.
+          if (!sessionStorage.getItem('FAKE_CYPRESS_LOGIN')) {
```
JazzarKarim, line 181: Nice to see that this worked 👍

TVWerdal: Yeah, thanks for this @JazzarKarim - got me passed a block.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-189-2jvfay4w.web.app

severinbeauvais (approved):
LGTM. The sample link works great!

JazzarKarim (approved):
Looks good.

PR #206, opened 2025-06-13
29002 update Add_Staff_Filing button menu

bobdev-94:
*Issue:*https://github.com/bcgov/entity/issues/29002 

*Description of changes:*

### DIFF
--- a/src/components/bcros/filing/addStaffFiling/Index.vue
+++ b/src/components/bcros/filing/addStaffFiling/Index.vue
@@ -220,9 +220,9 @@ const actions: ComputedRef<Array<Array<MenuActionItem>>> = computed(() => {
   const filteredActions =
     baseActions.length === 2
       ? baseActions.filter(
-        ({ datacy }) =>
-          datacy !== 'consent-to-amalgamate-out' &&
-          datacy !== 'consent-to-continue-out'
+        ({ label }) =>
+          label !== t('label.filing.staffFilingOptions.consentToAmalgamateOut') &&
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-206-wm75oeq8.web.app

src/components/bcros/filing/addStaffFiling/Index.vue
```diff
@@ -213,8 +213,20 @@ const allActions: ComputedRef<Array<MenuActionItem>> = computed(() => {
...
+  const filteredActions =
+    baseActions.length === 2
+      ? baseActions.filter(
+        ({ datacy }) =>
+          datacy !== 'consent-to-amalgamate-out' &&
+          datacy !== 'consent-to-continue-out'
+      )
+      : baseActions
+
+  return [filteredActions]
```
severinbeauvais, line 229: I agree that you have to filter these actions out of the list (since the list displays if num actions > 0).

However, I don't like the idea of checking a Cypress Test property to identify the actions to filter out. I suggest expanding MenuActionItem with a new property (of type AuthorizedActionsE maybe?).

Any other ideas?

bobdev-94: How about checking with the `label` not `datacy`?

severinbeauvais (approved):
LGTM.

Tested and works fine.

Wait for another review before merging.

TVWerdal (approved):
I see there is one unit test failing you may want to check on , but the code looks great and works well.

PR #209, opened 2025-06-24
29146 - Fixed apiURL is not defined bug

JazzarKarim:
*Issue:*https://github.com/bcgov/entity/issues/29146

*Description of changes:*
I'm so confused as to how this worked before. I fixed this bug:
image


bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-209-vdhgga3x.web.app

severinbeauvais (approved):
OK, this makes sense to me now. Approved.

I'll have one last look at the preview URL when it's updated.

severinbeauvais:
URL is OK but X-Apikey is missing:

image

JazzarKarim:
> URL is OK but X-Apikey is missing:
> 
> image

This is probably a temp link issue. We can probably fix that later.

severinbeauvais:
There are still issues but I've asked @bobdev-94 to fix them...

The following calls still use the old Legal API URL:
- business
- comments
- addresses
- parties

The following calls use the new Business API GW URL:
- allowable actions
- tasks
- filings
- polling (slim=true)

PR #212, opened 2025-06-26
29010 - Add Business API URL to vaults

JazzarKarim:
*Issue:*https://github.com/bcgov/entity/issues/29010 

*Description of changes:*


severinbeauvais:
/gcbrun

I know this will probably fail but I like to see the checkmark :)

bobdev-94 (approved):

JazzarKarim:
@bobdev-94 For context, this vaults file are where the keys are picked from when deploying to DEV, TEST, and PROD. If a value is not here, when building to DEV, it won't be populated and it will be empty. Those values are picked from 1pass.

TVWerdal (approved):

PR #213, opened 2025-06-27
28605 fetch authorized actions from legal api

TVWerdal:
*Issue:*https://github.com/bcgov/entity/issues/28605

- Updates dashboard ui so it uses legal-api to establish authorized actions. 
- Updates to cypress tests.


TVWerdal:
> What is this PR waiting on? Can it be expedited (ie, deployed to Prod within a few days)?

Still working on some of the cypress tests - have a way forward though I think.

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-213-ulpzife9.web.app

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-213-ulpzife9.web.app

meawong (approved):
Tested a some cases as Staff vs Regular user - LGTM!

cypress/e2e/components/comments.cy.ts
```diff
@@ -18,7 +19,10 @@ context('Business dashboard -> Comment side modal', () => {
   }
 
   it('Comment side modal is rendered', () => {
-    cy.visitBusinessDashFor('businessInfo/ben/active.json', undefined, false, false, undefined, allFilings, true)
+    // cy.interceptAuthorizedActions(BusinessRegistryStaffRoles).as('getAuthorizedActions')
```
severinbeauvais, line 22: Not needed? Delete?

src/composables/useBcrosAuth.ts
```diff
@@ -52,6 +52,9 @@ export const useBcrosAuth = () => {
           console.info('Checking account status...')
           // verify account status
           verifyAccountStatus()
+          // Get authorized Actions
+          console.info('Gathering user permissions...')
```
severinbeauvais, line 56: I'm in favour of not publicizing this. Thoughts?

src/stores/account.ts
```diff
@@ -48,14 +48,12 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
       throw new TypeError('Invalid roles')
     }
 
-    // FUTURE: when we fetch authorized actions from Legal API, we'll instead check
-    //         that the list of actions isn't empty
+    // Check that the list of actions isn't empty
     const allRoles = Object.values(AuthorizationRolesE)
     if (!allRoles.some(role => authRoles.value.includes(role))) {
```
severinbeauvais, line 53: This doesn't look right. These are the authorization roles (eg, STAFF), not the authorized actions. It may be enough to check that the authRoles is an array, or maybe validate that there's at least 1 role in there.

Eg, https://github.com/bcgov/business-create-ui/blob/92c5e1b245291931d95c8838188e6cd344b28f8c/src/App.vue#L1328

src/stores/account.ts
```diff
@@ -197,6 +206,29 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
...
   }
 
+  /**
+   * Load action called in useBcrosAuth.
+   * Uses the fetchAuthorizedActions to fetch the actions from the Legal API and set them in the store.
+   */
+  async function loadAuthorizedActions (): Promise<void> {
+    const authorizedActions = await fetchAuthorizedActions().catch(() => null)
+    // verify we have _some_ authorized actions
+    if (!Array.isArray(authorizedActions) || authorizedActions.length < 1) {
```
severinbeauvais, line 216: How is this different than line 81?

src/stores/account.ts
```diff
@@ -222,19 +254,42 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
...
     userFullName,
-    isStaffAccount,
     isCAAccount,
     accountErrors,
     activeProducts,
     updateAuthUserInfo,
     verifyAccountAuthorizations,
     setUserName,
     setAccountInfo,
+    setAuthorizedActions, // For testing purposes
```
severinbeauvais, line 290: For your testing purposes, or for some unit test somewhere?

src/composables/useBcrosAuth.ts
TVWerdal: I can remove it, just added it there to match the previous implementation.

severinbeauvais: Hmm, ok, leave it in :)

src/stores/account.ts
TVWerdal: You are right, I'll remove this

src/stores/account.ts
TVWerdal: loadAuthorizedActions uses fetchAuthorizedActions to make a call to the permissions endpoint. It is used to fetch the authorized actions and put them in the store. getAuthorizedActions is a quick call to the store to return the authorizedActions object itself for use in the app.

src/stores/account.ts
TVWerdal: For unit tests

cypress/e2e/components/comments.cy.ts
TVWerdal: Not needed - good catch.

src/stores/account.ts
TVWerdal: Should I maybe change the function name to 'setupAuthorizedActions' or something? Would that help to make it more clear?

src/composables/useBcrosAuth.ts
severinbeauvais: I see it's removed, lol.

I like consistency but I also didn't want to publicize that there's Permissions!

severinbeauvais:
Sorry, I merged this before any final changes. You can create another PR linked to the same ticket if there are more changes you want to make.

PR #220, opened 2025-08-15
30174 hotfix missing address loading fails-1

bobdev-94:
*Issue:*https://github.com/bcgov/entity/issues/30174

*Description of changes:*

Part -1: Business API GW calls

Fix function getBusinessAddress()
Fix other Business API GW calls that reference error.value?.data?.message -> error.value?.data?.rootCause?.message
Change app version to 1.1.11a

### DIFF
--- a/src/stores/account.ts
+++ b/src/stores/account.ts
@@ -259,7 +259,7 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
           console.warn('Error fetching authorized actions.', error.value)
           accountErrors.value.push({
             statusCode: error.value?.statusCode ?? StatusCodes.INTERNAL_SERVER_ERROR,
-            message: error.value?.data?.message,
+            message: error.value?.data?.rootCause?.message,
             category: ErrorCategoryE.ACCOUNT_ACCESS
           })
           return []
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.1.11.a",
+  "version": "1.1.11a",
   "scripts": {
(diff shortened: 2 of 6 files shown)


src/stores/account.ts
```diff
@@ -259,7 +259,7 @@ export const useBcrosAccount = defineStore('bcros/account', () => {
           console.warn('Error fetching authorized actions.', error.value)
           accountErrors.value.push({
             statusCode: error.value?.statusCode ?? StatusCodes.INTERNAL_SERVER_ERROR,
-            message: error.value?.data?.message,
+            message: error.value?.data?.rootCause?.message,
```
severinbeauvais, line 262: What about the Legal/Business API calls to fetch documents or comments (using URL from API)?

bobdev-94: - It's Post, not Get, so I am not sure what Api response for error
https://github.com/bcgov/business-dashboard-ui/blob/0fc463576497eff51a566d59a0766d596a2ce6e2/src/stores/business.ts#L477-L486

https://github.com/bcgov/business-dashboard-ui/blob/0fc463576497eff51a566d59a0766d596a2ce6e2/src/utils/filings.ts#L108-L118

- These are not problem. I think don't need update, because don't reference the xxx?.xxx?.message
https://github.com/bcgov/business-dashboard-ui/blob/0fc463576497eff51a566d59a0766d596a2ce6e2/src/stores/business.ts#L56-L68

https://github.com/bcgov/business-dashboard-ui/blob/0fc463576497eff51a566d59a0766d596a2ce6e2/src/utils/filings.ts#L48-L73

severinbeauvais: Sounds good. Resolving.

PR #222, opened 2025-08-22
30238 cleaned up some env keys + fixed breadcrumb

severinbeauvais:
*Issue:* bcgov/entity#20238

*Description of changes:*
- cleaned up sample env file
- deleted unused config from Nuxt file
- added Business Registry URL
- app version = 1.1.14
- updated browsers list (and pnpm lock file)
- removed unused keys from Vaults file
- used Business Registry URL in breadcrumbs
- fixed Cypress test


pnpm-lock.yaml
```diff
@@ -115,7 +115,7 @@ importers:
         specifier: ^8.57.0
         version: 8.57.0
       eslint-plugin-tailwindcss:
-        specifier: ^3.17.4
+        specifier: ^3.18.0
```
severinbeauvais, line 118: This and the changes below are the result of updating the browsers list (caniuse), which needs to be done from time to time.

src/utils/breadcrumbs.ts
```diff
@@ -12,7 +12,7 @@ export function getRegistryDashCrumb (): BreadcrumbI {
   const account = useBcrosAccount()
   return {
     text: ref(t('breadcrumb.registryDashboard')),
-    href: `${useRuntimeConfig().public.authWebURL}account/${account.currentAccount.id}/business`
+    href: `${useRuntimeConfig().public.businessRegistryURL}account/${account.currentAccount.id}`
```
severinbeauvais, line 15: Now, this link will go straight to the BRD (instead of Auth Web which redirects to the BRD if you're logged in).

image

PR #226, opened 2025-09-15
29843 transition todo UI

kialj876:
*Issue:*https://github.com/bcgov/entity/issues/

*Description of changes:*
- transition application todo

Screenshot 2025-09-15 at 5 16 39 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "bcros-business-dashboard",
   "private": true,
   "type": "module",
-  "version": "1.1.16",
+  "version": "1.1.17",
   "scripts": {
     "build": "nuxt generate",
     "build:local": "nuxt build",


src/components/bcros/todo/Item.vue
kialj876, line 1: - changed some of the specific AR referencing to be more generic
- tweaked the 'Due Date' spacing/alignment as requested by design
- added in BcrosI18Helper (needed for transition case)

cypress/e2e/components/todos/transition-application-filing.cy.ts
severinbeauvais, line 1: Please fix the GHA errors in this file, eg.

image

cypress/fixtures/todos/taskTA.json
```diff
@@ -0,0 +1,95 @@
+{
+  "tasks": [
+    {
+      "enabled": true,
```
severinbeauvais, line 4: Is it really possible to have 2 enabled tasks? (This one and the one on line 51.)

kialj876: Mihai was very clear about this - they can't block each other

severinbeauvais (commented):
Is there a concept of saving a Transition draft and resuming it later?

cypress/fixtures/todos/taskTA.json
severinbeauvais: OK!

kialj876:
> Is there a concept of saving a Transition draft and resuming it later?

There is, but it is not ready on the Transition Application UI side yet so I am leaving it alone for now. I messaged our PO because that is something that was missed previously so we'll have a ticket added to finish that flow through

src/utils/todo/action-functions.ts
```diff
@@ -20,6 +20,10 @@ export const doFileNow = (item: TodoItemI) => {
       goToEditUI(path)
       break
     }
+    case FilingTypes.TRANSITION: {
```
severinbeauvais, line 23: I think this may be a different Transition filing. Ask Mihai about it.

severinbeauvais: I would expect your filing name to be different:

image

kialj876: It is the same thing on the API so it should be the same here as well

severinbeauvais: I leave it with you. I just know that those enum values have been in place since early 2023 😛

severinbeauvais (approved):
LGTM. Thanks for the examples. Please wait for more reviews.

PR #232, opened 2025-09-26
30679 move the filingTypeToName function to the lang file

BrandonSharratt:
*Issue:*https://github.com/bcgov/entity/issues/30679

*Description of changes:*
Moved the function as it exists currently to being mostly supported by the lang file. When this gets upgraded to using the layers the stuff in this PR (at least the lang file portion) should be moved to layers.

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -410,11 +410,13 @@
       "on": "on",
       "effectiveAsOf": "EFFECTIVE as of",
       "filed": "FILED",
+      "pending": "PENDING",
       "filedAndPaid": "FILED AND PAID",
       "filedAndPending": "FILED AND PENDING",
       "paid": "PAID",
       "rejected": "REJECTED",
       "withdrawn": "WITHDRAWN",
+      "paymentCompleted": "PAYMENT COMPLETED",
       "futureEffectiveIncorporation": "FUTURE EFFECTIVE INCORPORATION",
       "futureEffectiveContinuation": "FUTURE EFFECTIVE",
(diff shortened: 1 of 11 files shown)


src/lang/en.json
```diff
@@ -562,6 +562,49 @@
...
+    "changeOfName": "Legal Name Change",
+    "changeOfOfficers": "Officer Change",
+    "changeOfRegistration": "Change of Registration",
+    "continuationIn": "Continuation Application",
+    "continuationAuthorization": "Continuation Authorization",
+    "conversion": "Record Conversion",
+    "correction": "Correction",
+    "courtOrder": "Court Order",
+    "dissolution": "Dissolution",
+    // TODO: dissolutionSubtypes
```
severinbeauvais, line 590: Would it make sense to handle `dissolutionTypeToName` here, now?

If not, please change "TODO" (which means to do in this PR) to "FUTURE" (which means later). Thx.

severinbeauvais:
Thanks for doing this!

Can you please resolve the build error, which is apparently breaking 3 checks?

When it's ready to merge, remember to update the app version. Thanks!

src/lang/en.json
BrandonSharratt: I added the types there (note that it was a todo before) I did not however handle the function as I wasn't sure of the difference between "Involuntary Dissolution" and "Failed to File"

bcregistry-sre:
Temporary Url for review: https://business-dashboard-dev--pr-232-uqrmd6lg.web.app

